PR #37, opened 2015-06-28
Updated README

### LINKED ISSUE #32: Easy way to install WikiBase?
To install WikiBase on malta, I had to do the following:
```Bash
~/WikipediaBase $ sudo pip install -r requirements.txt
~/WikipediaBase $ sudo python setup.py install
~/WikipediaBase $ sudo apt-get install python-gdbm
~/WikipediaBase $ sudo pip install sqlitedict
```
and only then would WikipediaBase work. Could this either be explicitly documented, or alternatively, could there be an automatic install script?

Thanks,

Michael

michaelsilver:
gdbm dependency and development install
closes #32


fakedrake:
Before I merge this just so we can be clear: `pip install -r requirements.txt` *should not* be required anymore. Since WikipediaBase is to be used as a library too from now on all dependencies should be maintained in setup.py. Setup tools will do the right thing and install any missing deps when you do `python setup.py {install,develop,...}`

Could you remove requirements.txt and the relevant documentation and check that it works correctly for you before I merge this.

fakedrake:
Before I merge this just so we can be clear: `pip install -r requirements.txt` *should not* be required anymore. Since WikipediaBase is to be used as a library too from now on all dependencies should be maintained in setup.py. Setup tools will do the right thing and install any missing deps when you do `python setup.py {install,develop,...}`

Could you remove requirements.txt and the relevant documentation and check that it works correctly for you before I merge this.

michaelsilver:
Sorry for the delay, @fakedrake. I've now checked and after adding `sqlitedict` to the `setup.py`, `python setup.py install` works without the `requirements.txt`, which I have now deleted. Updated the `README.md` per your requrest, and even included the caution message you wrote in #38.

PR #38, opened 2015-06-28
Updated docs that we need gdbm

fakedrake:
See issue #32


michaelsilver:
Can you make sure to include the "Development Version" instructions from pull request #37 even though you are getting rid of "Release Version" for reference sake?

PR #40, opened 2015-06-28
Changed string formatting for python 2 and 3 compatibility.

### LINKED ISSUE #25: Fix travis tests
Tests run on my machine make them run on travis too.

alvaromorales:
Fixes #25

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 python:
     - "2.7"
+    - "3.4"
 
 script:
     - python setup.py test
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 python:
     - "2.7"
-    - "3.4"
 
 script:
     - python setup.py test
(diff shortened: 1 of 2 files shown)


alvaromorales:
I thought the code was compatible with Python 3, per the "pull request guidelines" in the contributing guide. But after my build failed miserably I remembered CONTRIBUTING.md is just a stock file.

Anyway ... we're not there yet with Python 3 compatibility. I think the best thing to do right now is to just test on Python 2.7, and focus on getting to production.

Supporting two major versions of python sounds like a nightmare. When the day comes to upgrade to Python 3, we can just use the 2to3 tool.

If possible, @fakedrake, try to match your dev environment to the Infolab's: Ubuntu 14.04 and Python 2.7.

fakedrake:
Thanks alvaro, I am working on a mac or archlinux with python 2.7. Even though I do try to keep in mind that everyone else insterested in wikibase uses ubuntu sometimes I just don't get it right. If this becomes a real problem I will set up CI for wikibase on a csail machine along with travis.

Issue #41, opened 2015-06-29
Rendered Attributes Missing Space(s)


michaelsilver:
I have noticed that for some attribute mappings fetched with
```Python
>>> from wikipediabase.util import get_infobox, get_meta_infobox
>>> get_meta_infobox('<TEMPLATE_NAME>').rendered_keys()
```
the rendered attributes are missing spaces between words. For example:
```Python
>>> from wikipediabase.util import get_infobox,
get_meta_infobox
>>> get_meta_infobox('Template:Infobox weapon').rendered_keys()
{'origin': u'Place\xa0of\xa0origin',
'diameter': 'Diameter',
'traverse': 'Traverse',
'secondary_armament': 'Secondaryarmament',
'hilt_type': u'Hilt\xa0type',
'weight': 'Weight',
'max_range': u'Maximum\xa0firing\xa0range',
'guidance': 'Guidancesystem',
'armour': 'Armour',
'sights': 'Sights',
'number': u'Number\xa0built',
'height': 'Height',
'barrels': 'Barrels',
'rate': u'Rate\xa0of\xa0fire',
'caliber': 'Calibre',
'yield': u'Blast\xa0yield',
'range': u'Effective\xa0firing\xa0range',
'cartridge_weight': 'Shell weight',
'pw_ratio': 'Power/weight',
'speed': 'Speed',
'feed': u'Feed\xa0system',
'depth': 'Maximum depth',
'service': u'In\xa0service',
'production_date': 'Produced',
'altitude': 'Flight altitude',
'carriage': 'Carriage',
'crew': 'Crew',
'width': 'Width',
'primary_armament': 'Mainarmament',
'transport': 'Transport',
'suspension': 'Suspension',
'haft_type': u'Haft\xa0type',
'type': 'Type',
'detonation': 'Detonationmechanism',
'accuracy': 'Accuracy',
'engine': 'Engine',
'ceiling': 'Flight ceiling',
'designer': 'Designer',
'elevation': 'Elevation',
'propellant': 'Propellant',
'cartridge': 'Shell',
'blade_type': u'Blade\xa0type',
'used_by': u'Used\xa0by',
'wars': 'Wars',
'filling_weight': u'Warhead\xa0weight',
'boost': 'Boost time',
'recoil': 'Recoil',
'breech': 'Breech',
'sheath_type': 'Scabbard/sheath',
'variants': 'Variants',
'steering': 'Steeringsystem',
'manufacturer': 'Manufacturer',
[...]

Issue #39, opened 2015-06-28
Slow rendering of meta infoboxes


fakedrake:
Meta-infoboxes use urlopen directly instead of going through the caching fetcher. Add caching functionality for POST requests to the fetcher.

Found this one out due to crappy internet connection. Yay for overpriced and unpredictable ISP infrastructure!

Issue #35, opened 2015-06-28
Sometimes we get the infobox attribute above the one we were lookign for 


fakedrake:
@alvaromorales found this one on issue #25 but it's a separate issue. I have *seen* it but when I tried to rerun the tests it was gone. The problem is that the get query asks for the area of the sea and the max width (the previous attribute to the correct one) gets returned.

```
Traceback (most recent call last):
  File "/afs/csail.mit.edu/u/a/alvarom/chris-pywikipediabase/tests/test_resolvers.py", line 108, in test_compat_rx
    self.assertRegexpMatches(ans, rx, msg=msg)
AssertionError:
    Query: '(get "wikipedia-ocean" "Sea of Azov" (:CODE "DEPTH"))'
    Answer: '((:CODE "39,000Â km2 (15,000Â sqÂ mi)1"))'
    Matcher: '7.*met.*23.*ft'
    Completion: 29\38: '7.*met.*23.*ft' not found in u'((:CODE "39,000\xc2\xa0km2 (15,000\xc2\xa0sq\xc2\xa0mi)1"))'
```

fakedrake:
This appeared on travis too but travis now passes and every time I spotted it I then tried

    >>> from wikipediabase.util import get_infobox
    >>> get_infobox('sea of azov').get('area')
    >>> u'39,000\xc2\xa0km2 (15,000\xc2\xa0sq\xc2\xa0mi)[1]'

Which is the correct answer.

alvaromorales:
Note that the Travis build passed because my pull request modified the `.travis.yml` file. Travis recreates the test environment every time you do that (therefore deleting pages.db and renderer.mdb). I believe subsequent builds will fail, since pages.db is not being cleaned up.

Repro steps:
* I pulled the latest version of the code on an Infolab machine, and removed pages.db and renderer.mdb.
* Ran `python setup.py test`, all tests pass.
* Ran `python setup.py test` again. `test_compat_rx (tests.test_resolvers.TestResolvers)` fails on the Sea of Azov.
* Deleted renderer.mdb. test_compat_rx still fails.
* Deleted tests/data/pages.db. All tests pass.

Any ideas why / where in the code this problem is happening? Some context on this bug would be helpful for debugging.

fakedrake:
[...]

PR #43, opened 2015-06-29
Infobox keys contain no newlines or br tags

fakedrake:
Fixes issue #41 and issue #39


fakedrake:
This failed but I am merging because the cause is issue #35

Issue #46, opened 2015-07-29
Caching proposal


alvaromorales:
I've been getting familiar with how WikipediaBase does caching, and I have a few ideas on how to improve the current solution to make it production-ready. I would appreciate any feedback on this proposal.

### Summary

1. Move to Redis for caching source wikimarkup pages, HTML pages, rendered pages, and meta infoboxes
2. Remove Python object caching

### Redis

We need caching for two reasons: speed, and to reduce the number of requests to [wikipedia.org](https://en.wikipedia.org/wiki/Main_Page).

For those who don't know, [Redis](http://redis.io/) is an in-memory key value store. It is used by many websites for caching. It supports neat features like automatic key expiry, eviction policies, and data structures like sets or maps.

Since it runs in memory, it's very fast. If redis crashes, however, data is lost. We're using redis purely as a cache, not for storing any vital data. If we lose the cache, we simply bring redis back up and start with an empty cache (FWIW I'm not concerned by redis's stability since it's used by many companies in production).

I propose we replace gdbm with Redis due to 1) keys that automatically expire and 2) speed.

> Expires: Redis allows to set a time to live different for every key so that the key will be automatically removed from the server when it expires.

With auto-expiring keys we'll be able to mostly keep up with the latest version of Wikipedia.

Ideas for expiry policies:

* Source and rendered pages: 2-4 weeks
* Meta infoboxes: never, or something like 6 months

I would also like to introduce a hard dependency on Redis: WikipediaBase will need a Redis server running on localhost on port 6379 (default). Currently, some of the code assumes that there is a cache, some doesn't. I'd like to eliminate this added complexity. I believe that any intensive use of WikipediaBase will need caching anyway.

Our test server, Travis, supports starting a Redis container before the test suite run. We will be able to continue to use our continuous integration pipeline.

### Remove memoization

WikipediaBase tries to cache Python objects where possible (through `@memoize` and `_context_get` in `wikipediabase/util.py`). Here's a comment from the code:

```
    The context is actually a global cache used to keep track of the
    objects created and reuse them when possible.
    For example if I need an infobox for bill clinton I can get it by
    calling. _get_context('bill clinton', 'infoboxes' Infobox). which
    will create an Infobox instance for 'bill clinton' and cache
    it. If I later try to create an object in the same way, I will be
    reusing the old one.
    In place of symbol an object can be passed. In that case we return
    the object itself. This way we can have some flexibility with
[...]

Issue #31, opened 2015-06-24
Use Wikipedia API instead of Wikipedia Mirror?


michaelsilver:
I think we can at least partially replace using the Wikipedia Mirror with the [Wikipedia API](https://www.mediawiki.org/wiki/API:Main_page), at least when fetching individual articles. Play around with their [API Sandbox](https://en.wikipedia.org/wiki/Special:ApiSandbox#action=query&titles=Main%20Page&prop=revisions&rvprop=content&format=jsonfm) to get a sense of the features. Basically, you send a HTTP GET query with the base address `https://en.wikipedia.org/w/api.php?` and then list a bunch of parameters. You can get the result back in both in JSON or XML, your choice, just change `format=xml` to `format=json`. We shouldn't get banned, per their explanation on [API Ediquette](https://www.mediawiki.org/wiki/API:Etiquette):
>Request limit

>There is no hard and fast limit on read requests, but we ask that you be considerate and try not to take a site down. Most sysadmins reserve the right to unceremoniously block you if you do endanger the stability of their site.

>If you make your requests in series rather than in parallel (i.e. wait for the one request to finish before sending a new request, such that you're never making more than one request at the same time), then you should definitely be fine. Also try to combine things into one request where you can (e.g. use multiple titles in a titles parameter instead of making a new request for each title. Or use "generator" if possible instead of making a request for each result from another request).

Ex: To get Barack Obama's rendered Wikipedia page:
```
https://en.wikipedia.org/w/api.php?action=query&titles=Barack%20Obama&prop=revisions&rvprop=content&rvparse=&format=xml
```
Ex2: To get Barack Obama's unrendered (wiki markup) page:
```
https://en.wikipedia.org/w/api.php?action=query&titles=Barack%20Obama&prop=revisions&rvprop=content&format=xml
```
NB: the page titles need to be escaped properly (e.g. spaces are `%20`, etc.). Here is the Ruby WikiBase function to escape titles:
```Ruby
  def self.uri_escape(str)
    str = URI.escape(str)

    # Encode those characters missed by URI.escape because it was
    # treating them as special characters in the context of URIs
    h = {
      "$" => "%24",
      "&" => "%26",
      "+" => "%2B",
      "," => "%2C",
      "/" => "%2F",
      ":" => "%3A",
      ";" => "%3B",
      "=" => "%3D",
      "?" => "%3F",
      "@" => "%40"
    }
    h.each do |k,v|
      str.gsub!(k,v)
    end
[...]

Issue #21, opened 2015-03-30
Don't spam wikipedia.org for tests


fakedrake:
Turns out the admins at wikipedia.org will ban you from editing articles (ie using the sandbox feature) if you spam them. There are two solutions to this that should both be employed:

- Hit ashmore or another [wikipedia mirror](http://github.com/fakedrake/wikipedia-mirror)
- Use caches that for all the tests that can be version controlled so that the tests are really self contained. (eg. caching proxy server or a json saving fetcher)

PR #47, opened 2015-08-06
Fetch data from wikipedia.org

### LINKED ISSUE #44: Standardize Encodings
The inconsistencies in encoding are getting quite difficult to deal with. See #41 for an example -- `get_meta_infobox('<TEMPLATE_NAME>').rendered_keys()` is returning a mix of what are Unicode types (with `u` prefix), and normal string types (without the `u` prefix). And further, the newest change, as I mentioned in #41, is that the `u` prefixed strings are what actually should be normal strings that are encoded in utf-8. For example, `'rate': u'Rate\xa0of\xa0fire'` should actually be `'rate': 'Rate\xa0of\xa0fire'` (note the lack of `u`) since '\xc2\xa0' is a utf-8 encoded string. When decoded, it yields a Unicode string u'\xa0':
```Python
>>> 'Rate\xc2\xa0of\xc2\xa0fire'.decode('utf8')
u'Rate\xa0of\xa0fire'
```
Which then yields a proper Unicode string, which can be dealt with accordingly with `unidecode`:
```Python
>>> unidecode(u'Rate\xa0of\xa0fire')
'Rate of fire'
```

I propose that all strings are of the normal Python string type, just encoded in utf-8 (meaning `'Rate\xc2\xa0of\xc2\xa0fire'` instead of `u'Rate\xc2\xa0of\xc2\xa0fire'`). This way, no information is lost, such as when it is transliterated using `unidecode`, and in the case that someone wants to transliterate it, it is possible to use unidecode without unexpected errors (see #41).

I would fix this myself, but I am having trouble wrapping my mind around why things are encoded and decoded in the code, such as in `fetcher.py`, `persistentkv.py`, and `renderer.py`, and I have no idea how one managed to get `u'Rate\xc2\xa0of\xc2\xa0fire'` instead of `'Rate\xc2\xa0of\xc2\xa0fire'`. Thoughts?

alvaromorales:
Added a fetcher and renderer that fetches data directly from wikipedia.org (see #31). Implemented a CachingFetcher and CachingRenderer as proposed in #46.

I read the [FAQ](https://www.mediawiki.org/wiki/API:FAQ#get_the_content_of_a_page_.28wikitext.29.3F) on the Wikipedia API and ran experiments to determine the fastest way to get rendered and unrendered pages. I experimented with /index.php, /api.php, and RESTbase, and found that index.php was best for the Fetcher and api.php was best for the renderer.

Also fixed #44 -- all strings in WikipediaBase are now unicode, except when outputting an enchantment for START.

Unfortunately, this ended up being a *huge* change that broke several tests. I was unable to just swap out the implementation of the old WikipediaSiteFetcher because other abstractions like Article or Infobox made assumptions on the format of the data returned by the Fetcher. Wikipedia.org's data was different enough to break some regexes or XPATH expressions.

Apologies for the size of this PR, but I found some bugs along the way that needed to be fixed to get tests to pass.


alvaromorales:
Build failed because of #48. @fakedrake could you restart the build on Travis?

michaelsilver:
I'm surprised that you use the human-visitable, real Wikipedia website to get pages -- won't you get banned? I think @fakedrake said that when he last tried this he got banned. Or, do you have some sort of request limit in place that prevents that from happening?

I agree that we should not "push the limits" with the API, but the API is designed for this kind of stuff (computers are fast, whoever made the API realizes this).

alvaromorales:
Valid point; I'd be willing to switch back to using the Wikipedia API if we think it's better (that's what I first started working on). Especially if @fakedrake did indeed get banned.

As to why I chose /index.php:

1) I assumed that the same usage guidelines applied (serial requests are fine). 

2) The [API FAQ guidelines](https://www.mediawiki.org/wiki/API:FAQ#get_the_content_of_a_page_.28wikitext.29.3F) encourage it:

> If you just want the raw wikitext without any other information whatsoever, it's best to use index.php's action=raw mode instead of the API

> If you just want the HTML, it's best to use index.php's action=render mode instead of the API

3) I ran experiments and got better performance using index.php. 

The results below show the total time to fetch 41 articles.

```
Using /index.php
    source: 4.87s
    rendered: 4.87s

Using /api.php
    source: 5.40s
    rendered: 26.46s
```

fakedrake:
I did get banned from wikipedia but not entirely:
https://github.com/fakedrake/WikipediaBase/issues/21.

So we can't switch to wikipedia.org for renderer.py but I have no good
argument against using it for everything else unless we start crawling
wikipedia or something.

Another point that can be made is that renderer.py only renders meta
infoboxes. That means that it might be doable to have it cache every single
infobox type from ashmore and never have it access the internet again. Then
we use fast wikipedia.org for everything and everyone is (probably) happy.

Right now I am on vacation/business trip but I have been loosely working on
a mechanism to control global configuration at a single point (config
branch) so if everything works out as I picture it, eventually tests that
won't need to go out to the internet at all, and we will be able to decide
on what source the production should use whenever we want. And even switch
back and forth at runtime if needed.

On Sun, Aug 9, 2015 at 12:35 AM, Alvaro Morales <notifications@github.com>
wrote:

> Valid point; I'd be willing to switch back to using the Wikipedia API if
> we think it's better (that's what I first started working on). Especially
> if @fakedrake <https://github.com/fakedrake> did indeed get banned.
>
> As to why I chose /index.php:
>
> 1) I assumed that the same usage guidelines applied (serial requests are
> fine).
>
> 2) The API FAQ guidelines
> <https://www.mediawiki.org/wiki/API:FAQ#get_the_content_of_a_page_.28wikitext.29.3F>
> encourage it:
>
> If you just want the raw wikitext without any other information
> whatsoever, it's best to use index.php's action=raw mode instead of the API
>
> If you just want the HTML, it's best to use index.php's action=render mode
> instead of the API
>
> 3) I ran experiments and got better performance using index.php.
>
> The results below show the total time to fetch 41 articles.
>
> Using /index.php
>     source: 4.87s
>     rendered: 4.87s
>
> Using /api.php
>     source: 5.40s
>     rendered: 26.46s
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/fakedrake/WikipediaBase/pull/47#issuecomment-129046412>
> .
>



-- 
Chris Perivolaropoulos

alvaromorales:
Ah, yes the sandbox feature (editing articles) is more heavily restricted. But the renderer in this PR uses the Wikipedia API, which allows us to make serial requests (rendering included).

If it's ever a problem, though, I agree that we could fall back to ashmore.

Let me know when you get a chance to review this; I'll work on increasing test coverage next.

alvaromorales:
Just wanted to follow up on this. @fakedrake can you take a look when you have a sec?

fakedrake:
Yes sorry, I took a quick look at it as I am not at home. I trust that it's ok :)

alvaromorales:
Thanks! Feel free to assign me any issues in the future if you would like any changes. Hope you're enjoying the vacation/business trip :-)

PR #53, opened 2015-09-03
Remove all mentions of compatibility or classic modes

### LINKED ISSUE #50: Compatibility and classic modes
In "compatibility mode", WikipediaBase returns an answer with a typecode, e.g. `((:html "George W. Bush"))`. The opposite is "classic mode". "Compatibility mode" is the expected behavior of WikipediaBase, as implemented by Ruby-WikipediaBase.

As the code is currently written, every query to `knowledgebase.py` runs in "compatibility mode" (`_get` never looks at the value of the `compat` param).

Why do we need a "classic" mode? Perhaps switching everything to effectively run in "compatibility mode" could help us remove unnecessary complexity.

At least, the naming of "compatibility" and "classic" modes should be changed to something more descriptive such as `typed_answer=True`.

alvaromorales:
Compatibility and classic modes are gone -- WikipediaBase always returns answers with a typecode (this used to be 'compatibility mode'). Fixes #50.

Refactored resolvers to adhere to WikipediaBase naming conventions. The naming convention is for Wikipedia articles to be called `symbol` (except when dealing with an instance of the `Article` class). Infobox attributes should be called `attr` and Wikipedia classes `cls`.

Changed the `get` knowledgebase method now always require a wikipedia class.

### DIFF
--- a/tests/examples.py
+++ b/tests/examples.py
@@ -180,9 +180,9 @@ DEGENERATE_EXAMPLES = [
      '#t'),
     ('(get "wikipedia-term" "Purchasing power parity" (:calculated "PROPER"))',
      '#f'),
-    ('(get "Lamb of God (band)" "PROPER")',
+    ('(get "wikipedia-musical-artist" "Lamb of God (band)" "PROPER")',
      '#t'),
-    ('(get "Board game" "PROPER")',
+    ('(get "wikipedia-term" "Board game" "PROPER")',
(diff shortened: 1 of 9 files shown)


alvaromorales:
I'm going to merge this PR because it contains a fix to a unit test that is blocking other PRs. Let's go for a post-merge code review, please let me know if you would like any changes.

PR #59, opened 2015-09-24
Update badges in README and remove Travis email notifications

alvaromorales:
Remove build notification emails, they're spammy. Travis by default will send an email to the PR author if the build fails.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,3 @@ script:
     - coverage run setup.py test
 after_success:
     - coveralls
-notifications:
-  email:
-    - alvarom@csail.mit.edu
-    - cperivol@csail.mit.edu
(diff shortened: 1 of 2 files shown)


michaelsilver:
Looks good to me!

P.S. Waffle looks pretty cool. Should we integrate it into our workflow?

alvaromorales:
Thanks!

Regarding waffle: yeah, maybe. It looks like a lightweight tool on top of issues that might help with collaboration.

fakedrake:
I also liked waffle initially but I find it, like all kanban-like interfaces, to me more distracting than useful. I know other people like it though and waffle doesn't seem to break the rest of the issue system with weird tags etc.

PR #63, opened 2015-09-26
Improve attribute fetching in MetaInfobox and rename functions for consistency

### LINKED ISSUE #60: Improve infobox_scraper with get_best_page function
A MetaInfobox is an infobox of an infobox. It contains all valid attributes for an infobox. MetaInfoboxes are very important; they're used to get unrendered-rendered mappings, for example.

To get a list of valid attributes, we fetch template documentation pages ([example](https://en.wikipedia.org/wiki/Template:Infobox_officeholder)), and look for wikimarkup blocks like:

```
{{Infobox officeholder
 | honorific-prefix   = 
 | name               = 
 | native_name        = 
...
```

Because everything in Wikipedia is hard, it turns out that there are multiple sources for attributes, some better than others. In order of best to worst:

1. `<templatedata>` [block](https://www.mediawiki.org/wiki/Extension:TemplateData) inside a ` <script>` tag in documentation subpage, e.g. `Template:Infobox_officeholder/doc`
2. HTML in documentation subpage, e.g. `Template:Infobox_officeholder/doc`
3. HTML in documentation page, e.g. `Template:Infobox_officeholder`

As an added complication, redirects don't work for `Template:` pages. For instance, `Template:Infobox_french_commune` does not redirect to `Template:Infobox_French_commune` (note difference in capitalization).

A few tests broke because Wikipedia added a redirect from `Template:Infobox president` to `Template:Infobox officeholder`.

infobox_scraper needs to be enhanced and cleaned up:
* Follow redirects before attempting to fetch a template page
* Add a `get_best_page` function to select the best source of attributes

alvaromorales:
Improved and cleaned up MetaInfobox to get a more complete list of infobox attributes. Fixed unit tests that broke due to edit in Wikipedia. 

Also refactored some functions for consistency.
* Use 'attribute' to refer to an infobox attribute, instead of 'key'
* Standardize method names across Infobox, MetaInfobox, and Fetcher

Fixes #60


michaelsilver:
:+1: for improving readability/docs of the code. Before I give the green light to merge, I want to clarify what's been changed here. 

### Naming Changes
In addition to updating the unittests, you changed some naming conventions. WikiMap uses WikipediaBase as a library to get mappings from metainfobox. It used to do it in this way:
```Python
from wikipediabase.util import get_meta_infobox
get_meta_infobox('Template:Infobox foo bar').rendered_keys()
```
Is the following now the new way?
```Python
from wikipediabase.util import get_meta_infobox
get_meta_infobox('Template:Infobox foo bar').rendered_attributes()
```

### Behavior Changes
Has the output of `get_meta_infobox('Template:Infobox foo bar').rendered_attributes()` or however it is called changed in this PR (e.g. different or more mappings)? Could you give an example?

alvaromorales:
Yes, this is the new way:

```python
from wikipediabase.util import get_meta_infobox
get_meta_infobox('Template:Infobox foo bar').rendered_attributes()
```

I improved the logic to fetch attributes (`get_meta_infobox('Template:Infobox foo bar').attributes()`) We now get more attributes. In theory, we'll also get more rendered attributes. I don't have an example off the top of my head.

PR #64, opened 2015-09-28
Remove extra level of nesting for error enchantments

### LINKED ISSUE #61: Errors are enchanted with an extra level of nesting
This bug probably came from the WikipediaBaseBugList wiki page, where there's a syntax error in the sample test case:

```ruby
assert_equal('(((error attribute-value-not-found :reply "Currently alive")))',
    tcp('(get "wikipedia-person" "Bill Clinton" "DEATH-DATE")').strip,
    "Clinton is still alive")
```

The output should be `((error attribute-value-not-found :reply "Currently alive"))`.

Fun fact: There's a comment in the code about this. `# Useless 3ple parens needed.`

The extra paren is indeed useless!

alvaromorales:
Fixes #61


michaelsilver:
Looks good to me!

PR #65, opened 2015-09-28
Fix test that broke due to edit in Wikipedia

alvaromorales:


michaelsilver:
Holy guacamole you are fixing things fast!

Looks good to me :)

PR #67, opened 2015-09-28
Rename special classes to wikibase-class

### LINKED ISSUE #62: Rename special classes to wikibase-class
WikipediaBase generates special classes for symbols. `wikipedia-term` and `wikipedia-person` are examples. However, Wikipedia has a "person" infobox, so we have a name collision.

Let's rename special WikipediaBase-only classes to `wikibase-class` and call true Wikipedia classes `wikipedia-class`.

alvaromorales:
WikipediaBase generates special classes for symbols. wikipedia-term and wikipedia-person are examples. However, Wikipedia has a "person" infobox, so we have a name collision. Renamed special WikipediaBase-only classes to 'wikibase-class'. Only 'true' Wikipedia classes are called 'wikipedia-class'. Added 'wikibase-paragraphs' to StaticClassifier. Added unit test that checks for namespace collisions with infobox person.

Fixes #62

### DIFF
--- a/tests/examples.py
+++ b/tests/examples.py
@@ -98,26 +98,26 @@ WIKI_EXAMPLES = [
 DEGENERATE_EXAMPLES = [
     # DEGENERATE CASES
     # Resolved from text
-    ('(get "wikipedia-person" "Mary Shakespeare" "DEATH-DATE")',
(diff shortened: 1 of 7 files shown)


michaelsilver:
Looks good to me!

PR #72, opened 2015-10-01
Fix bugs in enchantments and add more tests

alvaromorales:
More improvements to enchantments after discussing with @sfelshin.

Instead of using Guile's #t/#f, use Common Lisp conventions: 't' for True and 'nil' for False.

Use :error for the error typecode.

Rename EnchantedStringDict to EnchantedDict, and allow it to have values other than strings.

Fix bugs in EnchantedString where unicode curly double quotes weren't correctly escaped.

### DIFF
--- a/wikipediabase/enchantments.py
+++ b/wikipediabase/enchantments.py
@@ -135,6 +135,7 @@ class EnchantedString(Enchanted):
     def val_str(self):
         v = re.sub(r"\[\d*\]", "", self.val)  # remove references, e.g. [1]
         v = re.sub(r"[[\]]", "", v)  # remove wikimarkup links, e.g. [[Ruby]]
+        v = output(unicode(v)) # unidecode
         v = v.replace('"', '\\"')  # escape double quotes
         v = u'"{0}"'.format(v)
         return v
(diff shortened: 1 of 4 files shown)


wikipediabase/enchantments.py
```diff
@@ -135,6 +135,7 @@ def typecode_str(self):
     def val_str(self):
         v = re.sub(r"\[\d*\]", "", self.val)  # remove references, e.g. [1]
         v = re.sub(r"[[\]]", "", v)  # remove wikimarkup links, e.g. [[Ruby]]
+        v = output(unicode(v)) # unidecode
```
michaelsilver: `unicode(v)` or `unidecode(v)` because your comment says "unidecode"?

michaelsilver:
LGTM!

wikipediabase/enchantments.py
alvaromorales: I updated the comment for clarity.

The `output` function in `util.py` is a single point where `unidecode` gets called. When we transition to returning unicode values, we can just change one line.

michaelsilver: Great! I like that design. PR LGTM.

PR #74, opened 2015-10-13
Support multiple infoboxes without conflicts and take classes into account for resolvers

### LINKED ISSUE #16: Take classes into account when looking for attributes in get queries
A get request may have the following syntax

    (get <wikipedia class> <symbol> <attribute>)

Some wikipedia symbols (pages) have more than one infoboxes of different types. These infoboxes may have overlapping tags. If the wikipedia class received from get corresponds to the type of one of the infoboxes, only use that infobox. This is only meaningful if there are overlapping properties in infoboxes.

### LINKED ISSUE #51: Take classes into account for get-attributes
Similar to #16, `get-attributes` needs to make use of the wikipedia class in the query. The following queries return the same attributes:

```
(get-attributes "wikipedia-officeholder" "Vladimir Putin")
(get-attributes "wikipedia-martial-artist" "Vladimir Putin")
```

The answer returned merges attributes in both types of infoboxes. But the `officeholder` and `martial-artist` infoboxes have some overlapping attributes such as `NAME` and `IMAGE`.

alvaromorales:
An article can have multiple infoboxes. For an example, see Vladimir Putin's article with infobox wikipedia-officeholder and wikipedia-martial-artist. Previously, all infoboxes were merged into one Infobox object. This causes namespace conflicts with attributes.

Changed the logic to scrape infoboxes. We now use InfoboxScraper to construct a list of Infobox objects. InfoboxScraper finds infoboxes in the markup source and rendered HTML. Renamed `get_infobox` to `get_infoboxes` in util.py.

Improved the Resolver abstraction. Resolvers should only return results when a set of conditions is met (via a `_should_resolve()` method and `@check_resolver` decorator).

Classes are now taken into account for knowledge base methods like `get` and `get-attributes`.

Fixes #16 and fixes #51


wikipediabase/resolvers/term.py
```diff
@@ -16,71 +16,41 @@ class StaticResolver(BaseResolver):
...
 
-        lat = src.find(".//span[@id='coordinates']//span[@class='latitude']")
-        lon = src.find(".//span[@id='coordinates']//span[@class='longitude']")
+            xpath = ".//span[@id='coordinates']"
+            lat = src.find(xpath + "//span[@class='latitude']")
+            lon = src.find(xpath + "//span[@class='longitude']")
 
-        if lat is None or lon is None:
-            return None
+            if lat is None or lon is None:
```
michaelsilver: The green highlight looks like something changed with whitespace -- just double checking that this is 4-space indenting.

setup.py
```diff
@@ -18,19 +18,20 @@
...
-        'overlay-parse',
+        'fuzzywuzzy',
+        'hiredis',
         'lxml',
-        'sqlitedict',
-        'requests',
-        'beautifulsoup4',
+        'overlay-parse',
         'redis',
         'redis',
```
michaelsilver: `redis` twice in the `setup.py`?

michaelsilver:
Wow, this is a big one! I added two comments on small things, but overall code looks great! :+1:

wikipediabase/resolvers/term.py
alvaromorales: I wrapped the block in a for loop, which explains the extra level of indentation.

alvaromorales:
Thanks for your comments, @michaelsilver!

PR #76, opened 2015-10-14
Remove pip caching on Travis and add requirements.txt

### LINKED ISSUE #75: Notice missing dependencies on Travis (pip caching)
It bugs me that Travis is caching pip dependencies. I really think, especially for PRs that make changes to the `setup.py`, Travis aught to be installing from scratch (w/o caching) to mimic real use-case. Otherwise, it's possible that we delete a dep and don't notice that it actually is still needed until someone does a fresh install somewhere else. This is relevant during development because WikipediaBase is currently being used as a library in other projects (e.g. WikiMap), and it should be up to the WikipediaBase testing to make sure all dependencies are present at install. Otherwise, we might be merging code that breaks things (by a missing dependency) and not noticing it in Travis.

Does pip caching ever expire? Maybe there is some way to cache always, *except for when setup.py is modified*, in which case we do a fresh install and create a new cache? Thoughts?

alvaromorales:
Closes #75

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ sudo: false
 language: python
 python:
     - "2.7"
-cache: pip
 services:
     - redis-server
 install:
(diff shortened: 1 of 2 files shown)


michaelsilver:
Looks good to me!

PR #79, opened 2015-10-20
Improve server and fix bugs for end-to-end testing

### LINKED ISSUE #83: (get "wikipedia-sea" "Black Sea" (:code "LENGTH" :rendered "Max. length")) --> malformed error
WikiBase is mishandling a `get` call where the attribute is specified with both `:code` and `:rendered` attribute names, and it's also mishandling the error it returns.  `get` must always return a list of values, each of which is a list.  E.g., `((:error ...))` but not `(:error ...)`. Note this bogus value:

> $ telnet nauru 8023
> (get "wikipedia-sea" "Black Sea" (:CODE "LENGTH" :RENDERED "Max. length"))
> (:error TypeError :message "lispify() got multiple values for keyword argument 'typecode'")

Regarding the `:code ... :rendered ...` issue, compare:

> $ telnet nauru 8023
> (get "wikipedia-sea" "Black Sea" (:code "LENGTH"))
> ((:html "1,175 km (730 mi)"))
> $ telnet nauru 8023
> (get "wikipedia-sea" "Black Sea" (:rendered "Max. length"))
> ((:error attribute-value-not-found :reply "Unknown"))

It's a bug that `get` fails with a rendered attribute name, but at least it fails with a well-formed error message.

alvaromorales:
Various bug fixes as we begin end-to-end testing.

Removed all calls to unidecode -- WikipediaBase should always return unicode.

Omnibase is running Guile 1.8, which doesn't support unicode characters. Because of this, we were converting all unicode strings to ascii strings using the unidecode library. It turns out that unidecode (converting unicode strings into normal byte strings) is only necessary when generating symbols and synonyms. It's safe for WikipediaBase to return answers in unicode, as START can handle them correctly.

Simplified the behavior of the telnet server by using Python's SocketServer library. Similar to Ruby-WikipediaBase, we're not binding sockets and are closing the connection after returning an answer.

Left some unfinished, lesser priority work as TODO comments.

End-to-end testing on nauru has begun! We're hopefully on track for our goal to launch WikipediaBase in production by December 2015.

EDIT: fixes #83

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,4 @@
 -e git://github.com/fakedrake/overlay_parse#egg=overlay-parse
-Unidecode==0.4.18
 beautifulsoup4==4.4.1
 coverage==4.0.1
 coveralls==1.0
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,6 @@ setup(
         'redis',
         'requests',
         'sqlitedict',
-        'unidecode',
         'unittest2 < 1.0.0',
     ],
     dependency_links=[
--- a/tests/test_lispify.py
+++ b/tests/test_lispify.py
@@ -25,7 +25,7 @@ class TestLispify(unittest.TestCase):
         self.assertEqual(lispify('foo \\ "bar"'), '"foo \\ \\"bar\\""')
 
     def test_unicode_string(self):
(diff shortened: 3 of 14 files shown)


michaelsilver:
LGTM

PR #99, opened 2015-11-29
Improve WikipediaBase performance by adding a backend

### LINKED ISSUE #70: Infobox attributes are deduplicated when rendering
Wikipedia deduplicates synonymous attributes before rendering an infobox. If both `SUCCESSOR` and `SUCCEEDING` are in the infobox to render, only one gets rendered. This means we lose a lot of unrendered-rendered mappings.

Example:

```python
In [1]: from wikipediabase.metainfobox import MetaInfobox

In [2]: m = MetaInfobox('Template:Infobox president')

In [3]: len(m.attributes())
Out[3]: 160

In [4]: len(m.rendered_attributes())
Out[4]: 69
```

We only get rendered mappings for less than 50% of attributes. Instead of rendering all attributes in one infobox:

```
{{Infobox president
| succeeded = foo
| succeeding = foo
...
}}
```

A naive solution would be to render each attribute in its own infobox, e.g.:

```
{{Infobox president
| succeeded = foo
}}

{{Infobox president
| succeeding = foo
}}
```

However, this render query takes a long time.

We should find a way to separate attributes into the smallest number of disjoint sets, perhaps by looking at prefixes of attributes.

EDIT: note that synonymous attributes might render to different things. `SUCCESSOR` renders to 'Successor' and `SUCCEEDING` renders to 'Succeeding'.

### LINKED ISSUE #85: Use Wikipedia dump to massively improve performance
### Performance analysis

I ran a performance test comparing WikipediaBase with Ruby-WikipediaBase. I collected about 400 random articles, and timed a `get-classes` query.

The gist: **WikipediaBase is 40x slower than Ruby-WikipediaBase**.

When all pages are cached in redis, WikipediaBase is 38x slower. When all pages are cached in redis and all Python objects are cached in memory, WikipediaBase is 2.3x faster.

There's a clear need for trying to optimize the code, as well as thinking about faster data sources.

I tried testing with the wikipedia mirror on ashmore, but getting the page for one article took 23 seconds on average. Extrapolating, this would make WikipediaBase about 3000x slower.

### Proposed solution

The goal is to make WikipediaBase **at least as fast** as Ruby-WikipediaBase.

After looking at the Ruby-WikipediaBase code and discussing with @sfelshin, we found that not all knowledge base methods need access to live rendered/unrendered pages. In fact, only `get` needs to hit a live page. `get-classes`, `get-attributes`, `sort-symbols`, and `sort-symbols-named` can just query a MySQL dump of Wikipedia with good indexes.

The plan is to build a well structured backend and update `DbFetcher`. Additionally, I will look for ways to optimize the code and lazy load objects where possible.

The MetaInfobox abstraction produces great unrendered-rendered mappings. Its implementation won't be modified.

### LINKED ISSUE #93: sort-symbols-named sorting backwards
(sort-symbols-named "beethoven" "Ludwig van Beethoven" "1815 Beethoven" "Beethoven (crater)" "Beethoven (film)" "Beethoven (film series)" "Beethoven (I Love to Listen To)" "Beethoven (movie)" "Beethoven (TV series)")
    ("1815 Beethoven" "Beethoven (crater)" "Beethoven (TV series)" "Beethoven (I Love to Listen To)" "Beethoven (film series)" "Beethoven (film)" "Beethoven (movie)" "Ludwig van Beethoven")

I haven't done a strict character count, but looking at several of these symbols' pages, it looks to me like `sort-symbols-named` is returning symbols ordered from shortest page to longest page. That's backwards of what it should do.  Compare

    (sort-symbols-named "ludwig van beethoven" "Ludwig van Beethoven" "1815 Beethoven" "Beethoven (crater)" "Beethoven (film)" "Beethoven (film series)" "Beethoven (I Love to Listen To)" "Beethoven (movie)" "Beethoven (TV series)")
    ("Ludwig van Beethoven" "1815 Beethoven" "Beethoven (crater)" "Beethoven (TV series)" "Beethoven (I Love to Listen To)" "Beethoven (film series)" "Beethoven (film)" "Beethoven (movie)")

which correctly returns the case-insensitive-exact-match symbol first, given that there is such a match.

alvaromorales:
This PR addresses the performance issues of WikipediaBase.

Added scripts to build the WikipediaBase backend. The backend has one table `articles` with three columns: `id, title, markup`. It uses the [peewee](http://docs.peewee-orm.com/en/latest/) ORM to abstract low-level SQL calls. Peewee is lightweight and the setup was trivial. It made things significantly easier than SQLAlchemy.

A main design goal with the backend was to make it as simple as possible. The simpler it is, the easier to rebuild and keep up to date. We're using Postgres as a backend, in an effort to use one database system for InfoLab services.

Added scripts to generate symbols and synonyms. The bulk of this work was translating from Ruby code. Fixed the bug with "bad" symbols that @michaelsilver documented in Ruby-WikipediaBase by recursively following redirects and disambiguations.

Building the backend from a Wikipedia dump takes about 10 hours. Generating symbols and synonyms takes 40 hours.

The main performance improvement was obtained by not having to fetch HTML for every request. We only need rendered HTML for `get`; we use the markup in the DB for `get-classes`, `get-attributes`, `sort-symbols`, and `sort-symbols-named`. Benchmarking shows that **WikipediaBase is at least as fast as Ruby-WikipediaBase** for most requests. In order to do this, we split infobox construction into two steps: `Infobox` only uses markup, and `RenderedInfobox` uses markup and HTML.

Did some cleanup of the code by removing circular dependencies between modules. Added symbols and synonyms files using Git Large File Service for Github. We're now going to be able to diff these files every time we rebuild the backend.

We're still going to be able to use Travis for CI. Instead of reading from a database, we default to fetch markup live from wikipedia.org when tests run on the Travis platform.

Apologies for the large PR -- adding a backend changed the code in many places. The good news is that Python-WikipediaBase is now running on kiribati, with a fresh backend and new symbols. We're starting a production test. We are on track for our goal to have WikipediaBase running in production by December.

Fixes #85 
Fixes #70 
Fixes #93


fakedrake:
this looks really nice.  it also makes my fork of Wikipediabase that mostly removed dependency injections (to ease testing and parametrization eg with us of api/scraping, wikipedia url, caching policy), separated concerns better ( eg cache related code from fetcher)

fakedrake:
sorry mistyped closing the pr


anyway  those issues seem less important in this change so thank you Alvaro.

wikipediabase/classifiers.py
```diff
@@ -36,49 +42,89 @@ def __call__(self, symbol, *av, **kw):
...
+        ".* person",
+        "^\d+ deaths.*",
+        "^\d+ births.*",
+        ".* actors",
+        ".* musicians",
+        ".* players",
+        ".* singers",
+    ]]
+
+    CATEGORY_MATCHES = [
```
michaelsilver: These seem sort of magical; are they infoboxes? Thinking ahead to a future maintainer: suppose Wikipedia changed, would I need to change these? and where should I look for new "categories"?

michaelsilver:
Overall, looks great. I especially really like what you did with `lispify` + `knowledgebase`. :+1: 

Some comments:

  1. I agree with @fakedrake -- it would be nice to update the README with instructions on how to fire up the backend and do all necessary set up for testing with WikipediaBase
  1. I see there was some serious refactoring in `infobox` -- has the command changed to get the attribute mappings? Last I heard, this is how to get the attribute mappings:
```python
from wikipediabase.util import get_meta_infobox
get_meta_infobox('Template:Infobox foo bar').rendered_attributes()
```
has this changed?

If you address these two things, the rest looks good to me! :rocket:

fakedrake:
Turns out you need `TRAVIS=true` for tests to work "out of the box". Otherwise you need to populate the postgres database. I think we need a better way of handling configurations like this (eg selecting online source (REST/html), selecting mirrors (ashmore/wikipedia.org/localhost/etc), switching backend databases, etc)

wikipediabase/classifiers.py
alvaromorales: These "categories" are actually Wikipedia article categories.

This method of classifying `wikibase-person` works surprisingly well. It was translated from Ruby-Wikipediabase. These regexes capture the most common categories assigned to people articles.

alvaromorales:
Hi all, thanks for the feedback! Sorry for disappearing for a while, I've been hosed with the end of the semester.

I updated the README with instructions on how to set up the backend.

The issues describe the fixes in detail. A summary:
- #85 was fixed by creating a postgresql backend and not fetching HTML for most requests
- #70 was fixed by separating attributes into the smallest number of disjoint sets, by looking at prefixes of attributes.
- #93 was fixed by reversing the sort order for sort-symbols-named

alvaromorales:
@fakedrake yes, if the env has `TRAVIS=true`, WikipediaBase will always fetch from live wikipedia.org instead of reading from the DB. `TRAVIS` is a special env variable set for all Travis machines.

I renamed it to something more sensible, `WIKIPEDIABASE_FORCE_LIVE=true`, and documented this in the README. Note the disclaimer that strongly recommends users/developers of WikipediaBase to set up the backend.

I'd like to avoid per-machine configuration as much as possible. For simplicity, WikipediaBase should work out of the box in any lab machine provided the backend was created. I don't think we need a way to handle configurations.

alvaromorales:
@michaelsilver, to get attribute mappings:

```python
from wikipediabase.infobox import get_meta_infobox
get_meta_infobox('Template:Infobox president').attributes

michaelsilver:
@alvaromorales all your changes look good to me! Thank you for updating the README, and also for updating me on the most recent version of the API call to get unrendered --> rendered mappings!

Issue #94, opened 2015-11-09
First-sentence date parsing should prefer precision to recall


sfelshin:
Currently Beethoven appears to have been born in the year 17:

    (get "wikibase-person" "Ludwig van Beethoven" (:CODE "BIRTH-DATE"))
    ((:yyyymmdd 00170000))

The first sentence for our dear friend is:

> Ludwig van Beethoven (Listeni/ˈlʊdvɪɡ væn ˈbeɪˌtoʊvən/; German: [ˈluːtvɪç fan ˈbeːtˌhoˑfn̩] ( listen); baptised 17 December 1770[1] – 26 March 1827)

Admittedly that's hard to parse, but if it's too hard, then `wikibase-person` should give up and return the date in `:html` format rather than `:yyyymmdd`, or it should give up entirely and throw some sort of error.  It should not claim that Beethoven could have had Christ for a baby sitter.

sfelshin:
Another example of bad date parsing:

    (get "wikibase-person" "Wolfgang Amadeus Mozart" (:code "BIRTH-DATE"))
    ((:yyyymmdd 00010000))

In this case the first sentence is

> Wolfgang Amadeus Mozart (German: [ˈvɔlfɡaŋ amaˈdeːʊs ˈmoːtsaʁt], English see fn.;[1] 27 January 1756 – 5 December 1791), ...

Perhaps the footnote mark is being mistaken for a date? It looks like the first-sentence date parsing is favors false positives over false negatives, and it should be the other way around, because the Infolab prioritizes precision over recall.

fakedrake:
Hello!
Indeed the problem is that it confuses the annotation for a date. There are actually two separate bugs that appear here. One is indeed the fact that  wikipediabase (see the very last 3 lines of resolvers/paragraph.py) which should be fixed. 

The other one is a problem with [overlay-parse](https://github.com/fakedrake/overlay_parse) which is a separate package that parses dates. It can parse individual dates and date ranges. Wikipedia-base first looks for a date range in the first parentheses. If that fails it falls back to the first date found which in this case is "1" ie "1 AD". The initial version of overlay-parse wasn't able to recognize unicode characters like the em dash so my best guess is that you have an old overlay-parse version installed. The blame for this problem is entirely on me. I hadn't registered it as a package on pypi. It there now and we can change setup.py to enforce `overlay-parse >= 1.1`.

alvaromorales:
[...]

PR #105, opened 2015-12-13
Whitelist text formatting tags in :html responses

### LINKED ISSUE #88: "km2" misindicates square kilometers
Here's what WikiBase currently returns:

    $ telnet nauru 8023
    (get "wikipedia-sea" "Black Sea" (:code "AREA"))
    ((:html "436,402&#160;km2 (168,500&#160;sq&#160;mi)"))

Here's the wikimarkup:

    | area               = {{convert|436402|km2|sqmi|-2|abbr=on}}

Here's the HTML source:

    <td>436,402&#160;km<sup>2</sup> (168,500&#160;sq&#160;mi)</td>

Shouldn't WikiBase return this?:

    $ telnet nauru 8023
    (get "wikipedia-sea" "Black Sea" (:code "AREA"))
    ((:html "436,402&#160;km<sup>2</sup> (168,500&#160;sq&#160;mi)"))

(Or alternatively `...436,402 sq km` or `...436,402 square kilometers`, but that seems like unhelpful complication of code.)

fakedrake:
Infoboxes clean up the html of the values removing all tags but a few. Added some text formatting tags to the whitelist.

Fixes #88

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+data/wikibase-term-*symbols filter=lfs diff=lfs merge=lfs -text
+data/wikibase-term-*synonyms filter=lfs diff=lfs merge=lfs -text
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,5 @@ script:
     - coverage run setup.py test
 after_success:
     - coveralls
+env:
+    - WIKIPEDIABASE_FORCE_LIVE=true
--- a/README.md
+++ b/README.md
@@ -65,6 +65,8 @@ Tests can then be run with:
 
     python setup.py test
 
+_Note: tests assume that you have set up the backend. To fetch from wikipedia.org instead of reading from the database, set the env variable `WIKIPEDIABASE_FORCE_LIVE=true`. This is strongly discouraged if you plan to use or contribute to WikipediaBase, as performance will be slow._
+
 ## API documentation
 
 Generate the documentation with:
--- a/README.md
+++ b/README.md
@@ -5,7 +5,6 @@
 
 [![Build Status](https://travis-ci.org/infolab-csail/WikipediaBase.svg?branch=master)](https://travis-ci.org/infolab-csail/WikipediaBase)
 [![Stories in Ready](https://badge.waffle.io/infolab-csail/WikipediaBase.svg?label=ready&title=Ready)](http://waffle.io/infolab-csail/WikipediaBase)
-[![# of downloads](https://pypip.in/d/wikipediabase/badge.png)](https://crate.io/packages/wikipediabase?version=latest)
 [![Coverage Status](https://coveralls.io/repos/infolab-csail/WikipediaBase/badge.svg?branch=master&service=github)](https://coveralls.io/github/infolab-csail/WikipediaBase?branch=master)
 
 ## Overview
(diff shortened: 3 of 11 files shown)


alvaromorales:
Looks good! Interesting fix.

> The reasoning behind the code there is that we remove all markup to make sure no weird stuff like this sneaks into the answer and whitelist only tags that we know are good, specifically li, ul and br

Do you have any examples of bad markup that might sneak it? I think we need a code comment that explains _why_ we're stripping markup, not just how. It should also explain why we whitelist some tags.

> If we go with the whitelist approach, other text formatting tags that probably should be whitelisted are `<b> <em> <i> <small> <strong> <sub> <sup> <ins> <del> <mark>`

This list looks pretty exhaustive to me, but I haven't checked the HTML spec. Do you think it's complete?

This change needs more tests. Have you tried `(get "wikipedia-president" "Bill Clinton" "RELIGION")`? Does the `[1]` reference get included?

alvaromorales:
The Beethoven and Mozart test cases were already passing previous to this change.

This PR proposes to pick the date / date range "with the least missing values". What does that mean exactly? Please give some examples of Wikipedia articles to understand how this improves precision.

I don't think this PR solves the underlying issue with dates in WikipediaBase, as described by @sfelshin in #94.

> It looks like the first-sentence date parsing favors false positives over false negatives, and it should be the other way around, because the Infolab prioritizes precision over recall.

WikipediaBase's current approach is to _always_ return :yyyymmdd for dates. START only uses the :yyymmdd answer for numeric calculations, e.g. "Was Mozart born before Beethoven?" For non-numeric calculations START is happy to show an :html answer.

For dates, instead of just returning:

```
(get "wikibase-person" "Ludwig van Beethoven" (:CODE "BIRTH-DATE"))
((:yyyymmdd 17701217))
```

we should return

```
(get "wikibase-person" "Ludwig van Beethoven" (:CODE "BIRTH-DATE"))
((:html "baptised 17 December 1770" :yyyymmdd 17701217))
```

For cases that are too hard to parse, we should avoid returning :yyyymmdd and only return :html. Is it possible to return :yyyymmdd only if we're absolutely confident?

fakedrake:
I agree about the tests. I committed some documentation about the tags. 

We should probably remove all annotations from infobox values as they are contextual and make a backlog issue to find a better way to handle them.

I got the element list from [w3schools](http://www.w3schools.com/html/html_formatting.asp). Since the site is educational and not comprehensive, and the html4 specification has information about text formatting [a](http://www.w3.org/TR/html401/present/graphics.html#h-15.2) [bit](http://www.w3.org/TR/html401/struct/text.html#h-9.2.1) spread I didn't provide a link in the comments.

alvaromorales:
I recently removed files stored on Git LFS from the repository. I'm going to have to close this PR because your fork references the old git history, and merging would dirty the history. I apologize for the inconvenience.

This change is almost almost there; I'll ask one of our students to add tests so we can merge this in.

alvaromorales:
I apologize -- I just noticed the latest commit with unit tests. This change is ready to be merged and looks good.

@cordeliaavery recently made changes to the HTML parsing code, which explains the merge conflicts. I'll ask her to de-conflict this patch and merge it in.

PR #107, opened 2015-12-13
Remove annotations before date parsing & prioritize

### LINKED ISSUE #94: First-sentence date parsing should prefer precision to recall

fakedrake:
Annotations can be mixed up with dates. Overlay parse is smart enough to
avoid this kind for dates but not so much with date ranges. Just remove
them before parsing.

Also prefer dates based on precision.

Fixes #94

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+data/wikibase-term-*symbols filter=lfs diff=lfs merge=lfs -text
+data/wikibase-term-*synonyms filter=lfs diff=lfs merge=lfs -text
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,5 @@ script:
     - coverage run setup.py test
 after_success:
     - coveralls
+env:
+    - WIKIPEDIABASE_FORCE_LIVE=true
--- a/README.md
+++ b/README.md
@@ -65,6 +65,8 @@ Tests can then be run with:
 
     python setup.py test
 
+_Note: tests assume that you have set up the backend. To fetch from wikipedia.org instead of reading from the database, set the env variable `WIKIPEDIABASE_FORCE_LIVE=true`. This is strongly discouraged if you plan to use or contribute to WikipediaBase, as performance will be slow._
+
 ## API documentation
 
 Generate the documentation with:
--- a/README.md
+++ b/README.md
@@ -5,7 +5,6 @@
 
 [![Build Status](https://travis-ci.org/infolab-csail/WikipediaBase.svg?branch=master)](https://travis-ci.org/infolab-csail/WikipediaBase)
 [![Stories in Ready](https://badge.waffle.io/infolab-csail/WikipediaBase.svg?label=ready&title=Ready)](http://waffle.io/infolab-csail/WikipediaBase)
-[![# of downloads](https://pypip.in/d/wikipediabase/badge.png)](https://crate.io/packages/wikipediabase?version=latest)
 [![Coverage Status](https://coveralls.io/repos/infolab-csail/WikipediaBase/badge.svg?branch=master&service=github)](https://coveralls.io/github/infolab-csail/WikipediaBase?branch=master)
 
 ## Overview
@@ -14,10 +13,14 @@ Wikipedia backend interface for
 [InfoLab's START](http://start.mit.edu). It answers s-expression
 queries via Telnet.
 
-## Usage
-Install `pip` and `gdbm` if not installed:
+## Documentation
 
-    apt-get install python-setuptools python-gdbm
+[API documentation](http://wikipediabase.rtfd.org)
+
+## Installing
+Install `pip` and `redis-server` if not installed:
+
+    apt-get install python-setuptools redis-server
 
 > Caution: Until wikipedibase is deployed on START installing it for non-testing purposes is not something one is supposed to do.
 
@@ -25,9 +28,31 @@ Install `wikipediabase`:
 
     python setup.py install
 
-## Documentation
+## Setting up the backend
 
-[API Documentation](http://wikipediabase.rtfd.org)
+1. Configure the backend. The scripts checks that Postgres and Redis are running, and creates the necessary tables.
+
+   ```
+   ./scripts/configure_db.py
+   ```
+
+   If Postgres throws `ERROR: new encoding (UTF8) is incompatible with the encoding of the template database (SQL_ASCII)`, follow [these](https://gist.github.com/ffmike/877447) instructions.
+
+1. Download the latest Wikipedia [dump](http://dumps.wikimedia.org/enwiki/).
+
+1. Create the backend. This step takes about 10 hours.
+
+   ```
+   bzcat /path/to/enwiki-YYYYMMDD-pages-articles.xml.bz2 | ./scripts/create_db.py
+   ```
+
+1. Generate symbols and synonyms (optional). This step takes about 40 hours.
+
+   ```
+   ./scripts/generate_symbols_and_synonyms.py
+   ```
+
+   Install [Git LFS](https://git-lfs.github.com/) and commit the changes to files in `data/`.
 
 ## Testing
 
--- a/data/wikibase-term-generated-synonyms
+++ b/data/wikibase-term-generated-synonyms
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:e4151c1f626936976e4be95e1c29d5f93d271decb45bea7434bc8d58d9108bc5
+size 563914871
--- a/data/wikibase-term-symbols
+++ b/data/wikibase-term-symbols
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:5073dbb0e14ad5b82623a01372cf8f001f282393f9d291c3ea9e1d1c2a97b898
+size 97023360
--- a/scripts/generate_symbols_and_synonyms.py
+++ b/scripts/generate_symbols_and_synonyms.py
@@ -19,8 +19,8 @@ logging.basicConfig(filename='symbols.log', level=logging.DEBUG)
 logging.getLogger("peewee").setLevel(logging.WARNING)
 
 # TODO: change paths; pass in data_dir in sys.argv
-symbols_f = open('../data/wikibase-term-foo-symbols', 'w')
-synonyms_f = open('../data/wikibase-term-foo-generated-synonyms', 'w')
+symbols_f = open('../data/wikibase-term-symbols', 'w')
+synonyms_f = open('../data/wikibase-term-generated-synonyms', 'w')
 
 latest_symbols = LRUCache(100)
 
--- a/tests/examples.py
+++ b/tests/examples.py
@@ -224,6 +224,11 @@ DEGENERATE_EXAMPLES = [
     # infobox defined outside the article, in its own template
     ('(get "wikipedia-military-conflict" "World War I" (:code "DATE"))',
      ('((:yyyymmdd 19140728))')),
+
+    ('(get "wikibase-person" "Wolfgang Amadeus Mozart" (:code "BIRTH-DATE"))',
+     '((:yyyymmdd 17560127))'),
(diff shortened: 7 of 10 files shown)


alvaromorales:
The Beethoven and Mozart test cases were already passing previous to this change.

This PR proposes to pick the date / date range "with the least missing values". What does that mean exactly? Please give some examples of Wikipedia articles to understand how this improves precision.

I don't think this PR solves the underlying issue with dates in WikipediaBase, as described by @sfelshin in #94.

> It looks like the first-sentence date parsing favors false positives over false negatives, and it should be the other way around, because the Infolab prioritizes precision over recall.

WikipediaBase's current approach is to _always_ return :yyyymmdd for dates. START only uses the :yyymmdd answer for numeric calculations, e.g. "Was Mozart born before Beethoven?" For non-numeric calculations START is happy to show an :html answer.

For dates, instead of just returning:

```
(get "wikibase-person" "Ludwig van Beethoven" (:CODE "BIRTH-DATE"))
((:yyyymmdd 17701217))
```

we should return

```
(get "wikibase-person" "Ludwig van Beethoven" (:CODE "BIRTH-DATE"))
((:html "baptised 17 December 1770" :yyyymmdd 17701217))
```

For cases that are too hard to parse, we should avoid returning :yyyymmdd and only return :html. Is it possible to return :yyyymmdd only if we're absolutely confident?

fakedrake:
Can start handle `((:html "val" :yyyymmdd num))` properly? I was under the impression that it was to be implemented.

overlay_parse can provide the text in which the dates it found. Would it be acceptable to provide `:yyyymmdd` only if the entire first parentheses of the text was considered in finding birth and death date, ie there is no extra context that would change the meaning of the dates found?

sfelshin:
On 12/16/15 6:15 AM, Chris Perivolaropoulos wrote:
>
> Can start handle |((:html "val" :yyyymmdd num))| properly? I was under 
> the impression that it was to be implemented.
>

Yes, but implementing it isn't much of a priority as long as 
WikipediaBase doesn't return it.  When WikipediaBase supports the 
behavior, I'll be delighted to make START use it.

> overlay_parse can provide the text in which the dates it found. Would 
> it be acceptable to provide |:yyyymmdd| only if the entire first 
> parentheses of the text was considered in finding birth and death 
> date, ie there is no extra context that would change the meaning of 
> the dates found?
>

That sounds like a promising tactic.

fakedrake:
>> overlay_parse can provide the text in which the dates it found. Would 
>> it be acceptable to provide |:yyyymmdd| only if the entire first 
>> parentheses of the text was considered in finding birth and death 
>> date, ie there is no extra context that would change the meaning of 
>> the dates found?
>>
> That sounds like a promising tactic.
On second thought I think a better way of dealing with this is to teach overlay_parse (instead of wikipedia base) to recognize lifespans instead of just date ranges (see below), that take up an entire sentence within the first parentheses. Otherwise we will be missing obvious cases like [Rosa Luxemburg](https://en.wikipedia.org/wiki/Rosa_Luxemburg):

> Rosa Luxemburg (also Rozalia Luxenburg; Polish: Róża Luksemburg; 5 March 1871[1] – 15 January 1919) ...

Or [Slavoj Žižek](https://en.wikipedia.org/wiki/Slavoj_Žižek)

> Slavoj Žižek (Slovene pronunciation: [ˈslavoj ˈʒiʒɛk] ( listen); born 21 March 1949) ...

To be a bit more precise:

- The text of interest should constitute an entire sentence.
- The text of interest is entirely within the first parentheses.
- The text of interest can be one of:
  - A time span, eg. *5 March 1871[1] – 15 January 1919*
  - A verb in simple past and then a date or date range 
    - If the verb is not *born* or *died* then consider the date accordingly
    - If the verb is something else we have [Beethowen](https://en.wikipedia.org/wiki/Ludwig_van_Beethoven)'s case where the date is useful information for the user but we can't trust to use it in parsed form.

Where text of interest is the minimum substring of the article from which we can infer the lifespan of a person.

Note: The current implementation just tries to induce the lifespan from the first date/date range it encounters.

alvaromorales:
I like this new approach!

> If the verb is not born or died then consider the date accordingly

What would the output be for Richard Shakespeare's dates? `Richard Shakespeare (1490–d. before 10 February 1561) ...`

Alas, because Everything In Wikipedia Is Hard, the text of interest _isn't_ always enclosed within the first parentheses. For an example of an article where the birth and death dates are not enclosed in parens, see [Edward Salim Michael](https://en.wikipedia.org/wiki/Edward_Salim_Michael).

> Edward Salim Michael was born in Manchester, England in 1921 and died near Nice, France in 2006.

This might just be a badly written article that doesn't follow Wikipedia conventions. If so, I say we should ignore it for the sake of precision.

fakedrake:
>> If the verb is ~~not~~ born or died then consider the date accordingly

Sorry that was a typo.

Yes I also think we would be risking precision if we tried to parse ESM's article correctly. 

As for Richard Shakespeare's death date, `overlay_parse`'s representation of dates, ie `(year : int, month : int, day : int)`, is not general enough to represent *before 10 February 1561*. It would be interesting to find out how many articles contain dates in the first parentheses but no valid life span can be induced by the forementioned rules. I expect it's not worth fixing, but I can't support this claim.

alvaromorales:
It would be great if, for the Richard Shakespeare case, we could just return:

```
(get "wikibase-person" "Richard Shakespeare" (:CODE "DEATH-DATE"))
((:html "d. before 10 February 1561"))
```

If overlay_parse isn't confident in it's parse, just default to HTML.

fakedrake:
Just to provide an update on this. What @alvaromorales suggests as the correct behavior I find very reasonable, but `overlay_parse` can't do this kind of thing the way it is structured right now. In the spirit of doing this correctly I am rewriting some parts of it to make it happen. Here is how I intend the whole thing to work.

If the information about a birth or death date is available in the article the possible responses by WikipediaBase should be either

- `:yyyymmdd` formatted date
- `:html` of the text where the relevant information can be found. The reason for prefering this is either because the date is unrepresentable in the `:yyyymmdd` format (eg. `mid 16th century`) or because contextual information that would make the `:yyyymmdd` format unreliable (eg. `baptised 17 December 1770`)

# Definitions

- All dates that can be evaluated to a `:yyyymmdd` values I will refer to as *"certain dates"* even if parts of the date are missing, eg `July 1345`, `22nd of June 1948`, `22nd of November`, `11/2/1987`.
- Dates for which WikipediaBase is expected to return non `:yyyymmdd` (eg `:html`) responses are *"qualified dates"*. These dates either can't be expressed in a `:yyyymmdd` fashion (eg. `mid 16th century`) or have a context that makes them misleading out of context (eg. `baptised 17 December 1770`). Qualified dates are generally very hard to parse or identify on their own.
- A *half date range* is a range with only a start or an end. The bound available must be a certain date. Eg. `1927-`, `22nd February 1927 to`, `until 3 BC`. Also note that `1927 to 1952` contains two half ranges.
- A  *"full date range"* is a date range both of which's bounds are certain dates.

# Heuristics

With these building blocks we can probably define birth and death dates more precisely using the following criteria in order.

- Birth and death found in the first parentheses take precedence over others.

- Birth dates and death dates are the lower and upper bound of full date ranges, that is constitutes the entirety of a sentence.

> Arthur Schopenhauer (German: [Ààa Åt ä Å Àà É…îp…ônÀåha äÃØ…ê]; **22 February 1788** ‚Äì **21 September 1860**)

- Birth and death dates are certain dates that are preceded by the word "born" or "died" accordingly and maybe a preposition (in/on).

> Jon Elster (born **February 22, 1940**, Oslo)

- Death dates are the upper bound of half date ranges that end a sentence (see [Beethoven's case](https://en.wikipedia.org/wiki/Ludwig_van_Beethoven)

> Ludwig van Beethoven (Listeni/?l?dv?? væn ?be??to?v?n/, /?be?t?ho?v?n/; German: [?lu?tv?ç fan ?be?t?ho?fn?] ( listen); baptised 17 December 1770[1] – **26 March 1827**)

- Birth dates are the lower bound of half date ranges that start a sentence (see [Richard Shakespeare's case](https://en.wikipedia.org/wiki/Richard_Shakespeare) )

> Richard Shakespeare (**1490**–d. before 10 February 1561)

- Death dates are the part of the sentence following a half range whose lower bound is a birth date. Proportionally birth dates are the part of the sentence preceding a half date range whose upper bound is a death date. This rule is the complementary of the previous two.


> Richard Shakespeare (1490–**d. before 10 February 1561**)


> Ludwig van Beethoven (Listeni/?l?dv?? væn ?be??to?v?n/, /?be?t?ho?v?n/; German: [?lu?tv?ç fan ?be?t?ho?fn?] ( listen); **baptised 17 December 1770[1]** – 26 March 1827)

# Missing cases

## Places in dates

It is fairly common that along with the dates the article make reference to the place where someone died/was born. This can complicate things ([Bernard-Joseph Saurin](https://en.wikipedia.org/wiki/Bernard-Joseph_Saurin)). An obvious way to get around this is to redefine certain dates to be `{date} {place separator} {rest of the sentence until date range separator}`. `{place separator}` would be `{"in", ","}`. This way these would qualify as certain dates. 

> 17 November 1781 in Paris

> February 22, 1940, Oslo

I am a bit skeptical as to how this would affect recall however because we just assume that a chunk of text is a place based on the preosition, or, even worse, a comma.

## Ranges as certain dates

> Jesus (/?d?i?z?s/; Greek: ?????? Iesous; 7–2 BC to AD 30–33)

Will make the above heuristics parse `33 AD => (:yyyymmdd -00330000)` as a certain death date and `7–2 BC to AD 30 => (:html "7–2 BC to AD 30")` as the birth date.

WikipediaBase in general will deal with Jesus' birth/death date correctly by prioritizing the infobox, but this is still a big weakness of the above proposal.

@sfelshin I feel like this could have been researched more rigorously than little me thinking about it really hard for a couple of hours. Google scholar and libgen didn't yield anything too useful, is there any bibliography that you know of that I can look into before continuing?

Any feedback is more than welcome.

sfelshin:
Just a quick comment for now:  Remember that precision is more important to us than recall, the perfect is the enemy of the good, and we prefer incremental rollout of functionality to a Really Big Idea that takes a really long time to implement.

As far as WikipediaBase is concerned, that means that the first priority is to implement a version of birth and death date parsing that only returns :yyyymmdd for dates in the first parenthesis in cases where it's easy to be sure they're unqualified, certain birth or death dates.  The second priority is to return :html otherwise.  The third priority is to return :html always, even when returning :yyyymmdd too.  Only after that would we like to incrementally improve recognition of unqualified date formats.

As for recognizing dates of any kind outside of the first parenthesis, before spending any time on this we should collect statistics on how common this is and read Wikipedia documentation for writing articles about people.  If this format violates Wikipedia conventions and is rare, then we don't care because GIGO.  If it's widespread and includes a lot of really famous people, then we might consider detecting it even if it violates Wikipedia conventions, but we might still decide to classify it as GIGO.  Rather than detecting the format, we might prefer to fix a few Wikipedia pages ourselves, or raise the issue with Wikipedia to see if they can encourage some of their contributors to fix the problem (providing Wikipedia with a list of bad pages might speed that up).

alvaromorales:
I recently removed files stored on Git LFS from the repository. I'm going to have to close this PR because your fork references the old git history, and merging would dirty the history. I apologize for the inconvenience.

PR #115, opened 2016-01-04
Check for pep8 style errors on every Travis build

alvaromorales:
Ignore E501, line is too long. Source files should be max 80 characters, but because of Python indentation this sometimes gets tricky. Instead of breaking the build with long lines, code reviewers should enforce this.

Fixed pep8 style errors and commented out a failing test.


alvaromorales:
@michaelsilver let's go for the post-merge review on this one; I'm trying to get the repo ready for new UROP students.

michaelsilver:
Just a sanity check: even though coveralls runs after pep8, only `coverage run setup.py test` is actually being measured for coverage? (i.e. running pep8 isn't counted in the coverage report). I think we're okay, since coverage didn't increase that much.

Looks good to me! :rocket:

PR #116, opened 2016-01-04
Add exclude patterns to correctly filter out non-people

### LINKED ISSUE #113: "The Beatles (TV series)" is not a person so why is it a wikibase-person?
WikipediaBase thinks it's a `wikibase-person`:

    (get-classes "The Beatles (TV series)")
    ("wikibase-term" "wikibase-sections" "wikipedia-television" "wikibase-person")

cordeliaavery:
Add exclude patterns to correctly filter out non-people
Include deities as people

Fixes #113


michaelsilver:
Awesome job! The code looks good; my only comment is that the tests seem a bit sparse -- could you think of some other examples other than "God" for deities that should be people? How about some more TV shows (or maybe also albums) that could have the same problem as "The Beatles (TV series)"? If your code still passes those tests, great. If not, maybe there are some more categories to exclude other than "based on". The example in issue #113 could perhaps indicate a larger problem. No need to go crazy looking for examples, but one test per each seems a little skimpy to me. If you find one more, perhaps a bit different example for each, I'd be satisfied (bonus: if you find an example that fails, and then add something to the code to make it even better at classifying!).

cordeliaavery:
the incorrect classification actually seems to be a larger problem. in general, it looks like a larger number of comic strips are misclassified; it seems common practice to give them the category "Fictional American people", "Fictional Swedish people", etc;

https://en.wikipedia.org/wiki/Grandma_(comic_strip)
https://en.wikipedia.org/wiki/Rocky_(comic_strip)

 is the "people" match in the regex necessary? it looks like if I take it out everything is still classified correctly, including the comic strips

sfelshin:
As for additional test cases, we should have:

   * a real person without birth or death dates (see https://github.com/infolab-csail/WikipediaBase/issues/96)
   * a mythical person with infobox (e.g., Methuselah)
   * a mythical person without infobox (e.g., Sarah)
   * a real person whose gender identity is neither male nor female
   * a non-specific deity (e.g., God)
   * a specific deity (e.g., God in Christianity, Parvati)
   * both male and female test deities
   * a fictional character with birth date in infobox (e.g., James T. Kirk) 
   * a fictional character with birth date in first sentence (if possible; I haven't found any) 
   * both male and female fictional characters
   * a named real animal (e.g., Jumbo the elephant, Elsa the lioness)
   * a named, gendered, fictional or mythical animal-as-sentient-being (e.g., Bambi)
   * a named, gendered, fictional or mythical animal-as-animal (e.g., Rikki-Tikki-Tavi)
   * a named, gendered, fictional, sentient alien (preferably one that doesn't look basically like a human except with a funny nose or ears; e.g., Jabba the Hut or Groot rather than a Vulcan or a Klingon)
   * a named, non-gendered, fictional, sentient alien (if you can find one)
   * a robot or machine viewed as animate (e.g., Hal 9000, WALL-E)
   * a fictional, animate non-living thing (e.g., Lumiere from Beauty and the Beast)
   * an animate organization with a singular name (e.g., IBM)
   * a set of people with a plural name, e.g., the Beatles
   * conjoined twins (e.g., Chang and Eng)
   * an article with dates in parens in the first sentence, but the numbers aren't birth or death dates (e.g., Korean War)

Note that the test cases should make sure the articles do or don't have birth dates, death dates, infoboxes, etc., as expected, so that if the articles change, we'll know to look for replacement test articles.

For now, we don't much care about the edge cases.  All real people should get class `wikibase-person`.  Fictional and mythical people, non-humans viewed as gendered and/or animate (deities, sentient aliens, talking animals, animals viewed as animate, non-living things viewed as animate, etc.), and conjoined twins can go either way -- it's preferable to classify them as people if they have either gender or birth/death dates, but either way is okay regardless.  Other articles should not get class `wikibase-person`.

Later we'll think more about why we care if articles are for people or not.  Is this about gender?  Sentience?  Animacy?  Singular personhood as opposed to groups?  Do we need to have multiple classes and/or attributes to capture multiple distinctions?

alvaromorales:
That's a great set of tests! We should create a new test suite in test_classifiers.py, `TestPersonClassifier`. Each test case should be in its own method, should group similar examples, and the method name should be descriptive (e.g `test_named_real_animal`).

> is the "people" match in the regex necessary? it looks like if I take it out everything is still classified correctly, including the comic strips

Perhaps the "people" pattern is too broad. If it doesn't hurt our precision, we should remove it.

> Note that the test cases should make sure the articles do or don't have birth dates, death dates, infoboxes, etc., as expected, so that if the articles change, we'll know to look for replacement test articles.

Just like Ruby-Wikipediabase, we're classifying an article as a person by looking at Wikipedia categories. We don't look at the presence of birth/death dates.

cordeliaavery:
I think I've covered a good number of the test cases above, but you might want to give tests/test_classifiers.py a once-over to see if the classifications seem ok @sfelshin

michaelsilver:
Looks good to me! @sfelshin should also take a look before you merge.

alvaromorales:
@sfelshin, here's a direct link to the person classification tests: https://github.com/infolab-csail/WikipediaBase/blob/beatles_fix/tests/test_classifiers.py

Glad to see that we're thoroughly testing PersonClassifier!

sfelshin:
Please rephrase comments like
  # Fails b/c of "Fictional characters" category
to increase clarity.  E.g., it's not clear to me whether the tests belonging to the above comment are expected to fail because the code is supposed to notice the "Fictional characters" category and neglects to, or because the code does notice the "Fictional characters" category and therefore mistakenly thinks (for what reason?) that that means the article is for a person.

Otherwise, LGTM!

cordeliaavery:
@sfelshin I've rephrased the comments; if everything looks good to you i'll merge

michaelsilver:
No need for @sfelshin to re-review; comments LGTM! Go ahead and merge :+1:

sfelshin:
Please open one or more new issues for all categories of "expected failing" test cases that you added, so that they don't get dropped on the floor.

Issue #81, opened 2015-10-20
Check length of first paragraph for the SHORT-ARTICLE attribute


alvaromorales:
The expected behavior of the SHORT-ARTICLE attribute for wikibase-term is to return:

> The first paragraph of the article, or if the first paragraph is shorter than 350 characters, then returns the first paragraphs such that the sum of the rendered characters is at least 350.

We're currently returning the first paragraph without checking its length.

alvaromorales:
The short-article attribute is provided by TermResolver: https://github.com/infolab-csail/WikipediaBase/blob/master/wikipediabase/resolvers/term.py#L105

sfelshin:
On Oct 20, 2015, at 3:34 PM, Alvaro Morales <notifications@github.com> wrote:
> The expected behavior of the SHORT-ARTICLE attribute for wikibase-term is to return:
> 
> The first paragraph of the article, or if the first paragraph is shorter than 350 characters, then returns the first paragraphs such that the sum of the rendered characters is at least 350.

The correct behavior is more complicated: If the SHORT-ARTICLE value is too short, it should get extended with extra text, it should get truncated, and furthermore, the text shouldn't get truncated in the middle of a word, and the return value should be legal HTML (no unclosed elements).  Do we have test cases for all that?

IIRC, the SHORT-ARTICLE script is also supposed to truncate the first paragraph if it's really long (and also not in the middle of a word, and as legal HTML) -- check the obsolete ruby implementation for the maximum length for a first paragraph.

alvaromorales:
Interesting. That method specification is from the wiki page. I looked at the Ruby-WikipediaBase code, and there's no truncation step. If the first paragraph is shorter than 350 characters, it keeps adding paragraphs until the total length is > 350 characters. This behavior is what Sherry's PR implemented.

Of course, we can open a new PR and change this. However, I'm not sure I see the point of truncating text in the middle of a sentence. I think it would reduce answer legibility. Are there any sample questions where START truncates snippets?

We don't have a test case that checks whether the return value is legal HTML. @sherryy can you add this test in a new PR?

sfelshin:
On Jan 6, 2016, at 6:19 PM, Alvaro Morales <notifications@github.com> wrote:
[...]

PR #125, opened 2016-01-06
Enable and test short_article to be at least 350 chars

sherryy:
Enable and test short_article to be at least 350 chars

#81


wikipediabase/resolvers/term.py
```diff
@@ -109,8 +109,17 @@ def short_article(self, symbol, _):
         that the sum of the rendered characters is at least 350.
         """
 
-        # TODO: check if the first paragraph is shorter than 350 characters
         first_paragraph = get_article(symbol).first_paragraph(keep_html=True)
+        first_paragraph_text = get_article(symbol).first_paragraph()
```
alvaromorales: You're correctly calculating the character count from the paragraph plaintext, instead of using the paragraph HTML with a lot of `<tags>`. However, the code could be made more efficient. The `get_article`, `first_paragraph` and `paragraphs` methods are quite expensive.

You should save `get_article(symbol)` to a local variable. To get paragraph text, you can just call:

```python
from wikipediabase.util import fromstring, totext

article = get_article(symbol)
first_paragraph = article.first_paragraph(keep_html=True)
first_paragraph_text = totext(fromstring(first_paragraph))
```

alvaromorales:
Looks great! :ship:

alvaromorales:
fyi, you have to type "Fixes" or "Closes" before referencing the issue number for Github to close the issue.

PR #126, opened 2016-01-06
Add usage section to README

michaelsilver:
Includes:
  * how to call WikipediaBase from the shell
  * Lisp API summary and examples

I was going to do this but hesitated: now that we have WikipediaBase running in production, can we get rid of 
>*Under active development. Go no further unless you know what you are doing.*

Seems a little ominous and off-tone given the rest of the README, especially now that WikipediaBase provides most of  it's main functionality (now even documented in the README).


alvaromorales:
Yes, you should do the honors of removing the `active development` warning! After that, LGTM

PR #127, opened 2016-01-07
Can use invoke for build tasks

### LINKED ISSUE #118: Create tasks.py for build tasks using invoke
[Invoke](http://www.pyinvoke.org/) is a Python library for build tasks, similar to Rake in ruby.

We should create a `tasks.py` file with these tasks:

* `install`: check that you're on the master branch and all code is committed. If not, ask for user confirmation. Install wikipediabase to a shared dir: `python setup.py install --home=/infolab/share/bin`
* `lint`: run linter (`pep8 wikipediabase scripts tests --ignore=E501`) 
* `test`: run unit tests
* `travis`: run `lint` and `test`

chsamlee:
Fixes #118 .


requirements.txt
```diff
@@ -8,6 +8,7 @@ edn-format==0.5.9
 flake8==2.4.1
 fuzzywuzzy==0.7.0
 hiredis==0.2.0
+invoke==0.11.1
```
alvaromorales: You also need to add `invoke` to `install_requires` in [setup.py](https://github.com/infolab-csail/WikipediaBase/blob/master/setup.py) (but without the version number).

tasks.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@task
+def install():
+    """
+    Install wikipediabase to a shared directory.
+    """
+    branch_name = popen("git rev-parse --abbrev-ref HEAD").read()
+    deploy = False
+    if branch_name == "master":
```
alvaromorales: I tried running this from the master branch, and still got the message "Not on master branch".

tasks.py
```diff
@@ -0,0 +1,43 @@
...
+        deploy = True
+    else:
+        proceed = raw_input("Not on master branch? Proceed? [y/n]")
+        if proceed == 'Y' or proceed == 'y':
```
alvaromorales: `proceed.lower() == 'y':`

alvaromorales:
Great work! I left some inline comments.

alvaromorales:
Ah, I forgot one more thing. When running `invoke install`, even on the master branch, we should ask for a confirmation. The warning should be something like `You're about to install a new version of WikipediaBase on all InfoLab machines. Proceed?`.

I got this error when running `invoke install`:

```
unning install
Checking .pth file support in /infolab/share/bin/lib/python/
/scratch/alvarom/.virtualenvs/pywb/bin/python -E -c pass
TEST FAILED: /infolab/share/bin/lib/python/ does NOT support .pth files
error: bad install directory or PYTHONPATH

You are attempting to install a package to a directory that is not
on PYTHONPATH and which Python does not read ".pth" files from.  The
installation directory you specified (via --install-dir, --prefix, or
the distutils default setting) was:

    /infolab/share/bin/lib/python/

and your PYTHONPATH environment variable currently contains:

    ''

Here are some of your options for correcting the problem:

* You can choose a different installation directory, i.e., one that is
  on PYTHONPATH or supports .pth files

* You can add the installation directory to the PYTHONPATH environment
  variable.  (It must then also be on PYTHONPATH whenever you run
  Python and want to use the package(s) you are installing.)

* You can set up the installation directory to support ".pth" files by
  using one of the approaches described here:

  https://pythonhosted.org/setuptools/easy_install.html#custom-installation-locations

Please make the appropriate changes for your system and try again.
```

It's the first time we're trying to install python packages to `/infolab/share/bin`. I'll look into what needs to be done.

sfelshin:
On 1/8/16 12:09 PM, Alvaro Morales wrote:
>
> Ah, I forgot one more thing. When running |invoke install|, even on 
> the master branch, we should ask for a confirmation. The warning 
> should be something like |You're about to install a new version of 
> WikipediaBase on all InfoLab machines. Proceed?|.
>

We should think about how we should clarify the distinction (in code, 
comments, notifications, documentation, etc.) between the WikipediaBase 
server program and the WikipediaBase backend.  The server program gets 
installed in a single location, namely the shared directory 
/infolab/share/bin/ (which is in everybody's path if they set up their 
dotfiles correctly); it corresponds to some version of code in the 
WikipediaBase repository. The WikipediaBase backend gets installed 
separately on the local drive of any machine where we choose to install 
it; it corresponds to some particular download of Wikipedia, as fetched 
and munged by some version of code in the WikipediaBase repository.  A 
warning message like "|You're about to install a new version of 
WikipediaBase on all InfoLab machines. Proceed?|" is a little vague 
about whether the server, the backend, or both will get installed, and 
whether it will get  installed once or installed on each machine 
individually.

alvaromorales:
The `install` tasks needs to be modified slightly. 

* Check for uncommitted files
* Run `test` before completing `install`

tasks.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    run("pep8 wikipediabase scripts tests --ignore=E501")
+
+
+@task
+def test():
+    """
+    Run the unit tests.
+    """
+    run("nosetests")
```
alvaromorales: I think it's better to run `python setup.py test` instead of `nosetests` here

alvaromorales:
Great work! I added some inline comments. I really like how robust the `install` task will be.

alvaromorales:
To get this PR merged, I've temporarily removed the `install` task. We still have to figure out some details on the deployment process, but I want this change merged. The `lint`, `test`, and `travis` tasks will make the dev process significantly smoother.

I'll open a separate issue for the `install` task.

PR #131, opened 2016-01-11
get takes plist instead of str as argument

### LINKED ISSUE #124: Handle plist in arguments to get
A plist is an alternating list of keywords and values, e.g. `(:code "FOO" :rendered "BAR")`. This translates to a Python dictionary `{'code': 'FOO', 'rendered': 'BAR'}`.

WikiBase is mishandling a `get` call where the attribute is specified as a plist with both :code and :rendered attribute names.

```
(get "wikipedia-sea" "Black Sea" (:code "LENGTH" :rendered "Max. length"))
((:error TypeError :message "lispify() got multiple values for keyword argument 'typecode'"))
```

These three formats are valid:

```
(get "wikipedia-sea" "Black Sea" (:code "LENGTH"))                            ;; works
(get "wikipedia-sea" "Black Sea" "LENGTH")                                    ;; works
(get "wikipedia-sea" "Black Sea" (:code "LENGTH" :rendered "Max. length"))    ;; breaks
```

The third argument to get can either be a string or a plist of values.

`wikipediabase/provider.py` is in charge of parsing the lisp method call into a Python method call. After updating, the `get` method in wikipediabase/knowledgebase.py will need to be updated,

chsamlee:
Fixes #124


wikipediabase/frontend.py
```diff
@@ -50,9 +50,20 @@ def _eval(self, ls):
...
-        args = [self._eval(a) for a in ls[1:]]
-        ret = fn(*args)
+        if isinstance(ls[0], Symbol):
+            fn = self.get_callable(self._eval(ls[0]))
+            args = [self._eval(a) for a in ls[1:]]
+            ret = fn(*args)
+        elif isinstance(ls[0], Keyword):
+            # argument is a plist; convert into python dict
+            ret = {}
+            for i in xrange(len(ls) // 2):
```
alvaromorales: Nice use of integer division syntax (`//`),  I didn't know about it.

wikipediabase/resolvers/base.py
```diff
@@ -55,16 +55,16 @@ def resolve(self, cls, symbol, attr):
...
+            return custom_resolve(cls, symbol, attrs)
+
+        for attr in attrs.values():
+            if isinstance(attr, LispType):
+                attr = attr.val
+            else:
+                attr = attr
+            attr = attr.lower()
+            if attr in self._resources:
+                return self._resources[attr](symbol, attr)
```
alvaromorales: This needs a code comment to explain why we need to loop over `attrs.values()`. In our plist, some attrs might specify an attribute name (e.g `:code`) and some might just be extra options (e.g. `:size`).

We need to support calls like:

`(get "wikibase-term" "Flower" (:code "IMAGE-LIVE" :size "350"))`

Don't you need to pass the `attrs` dictionary instead of the `attr` value to `self._resources[attr]` function? I think you need this to access some other param, e.g. `:size`, in a resolver. You might want to change this to:

    return self._resources[attr](symbol, **attrs)

alvaromorales:
LGTM! :rocket: I'm really excited about this change

PR #134, opened 2016-01-11
Add parsing capabilities for aicraft type infobox

### LINKED ISSUE #98: Can't parse meta infobox HTML for 'infobox aircraft type'
```python
>>> m = MetaInfoboxBuilder('Template:Infobox aircraft type')
>>> m.rendered_attributes()
{}
```

The rendered HTML has a weird format that we're not parsing correctly:

```html
p><br />
! Role | {{{type}}} |-</p>
<p><br />
! Design group |&#160;!!!!!design group!!!!! |-</p>
<p>! Built by |&#160;!!!!!builder!!!!! |-</p>
<p>! Introduction |&#160;!!!!!introduction!!!!! |-</p>
<p><br /></p>
<p><br /></p>
<p><br />
! <span class="nowrap">Program cost</span> |&#160;!!!!!program cost!!!!! |-</p>
<p><br /></p>
<p>! <span class="nowrap">Developed into</span> |&#160;!!!!!developed into!!!!! |-</p>
<p><br />
! Role |&#160;!!!!!type!!!!! |- ! <span class="nowrap">National origin</span> |&#160;!!!!!national origin!!!!! |- ! Manufacturer |&#160;!!!!!manufacturer!!!!! |-</p>
<p>! Designer |&#160;!!!!!designer!!!!! |-</p>
<p>! First flight |&#160;!!!!!first flight!!!!! |-</p>
<p>! Introduction |&#160;!!!!!introduced!!!!! |- ! Retired |&#160;!!!!!retired!!!!! |- ! Status |&#160;!!!!!status!!!!! |- ! Primary&#160;users |&#160;!!!!!primary user!!!!!<br />
!!!!!more users!!!!! |- ! Produced |&#160;!!!!!produced!!!!! |- ! <span class="nowrap">Number built</span> |&#160;!!!!!number built!!!!! |-</p>
<p>! <span class="nowrap">Unit cost</span></p>
<p>|</p>
<div>!!!!!unit cost!!!!!</div>
<p>|- ! <span class="nowrap">Developed from</span> |&#160;!!!!!developed from!!!!! |- ! Variants |&#160;!!!!!variants with their own articles!!!!! |-</p>
<p><br /></p>
```

We should be able to parse this format, and extract many rendered attribute pairs like `"Built by" <--> "Builder"`.

Are we failing to parse other infobox templates?

cordeliaavery:
Fixes #98


tests/test_infobox.py
```diff
@@ -138,6 +138,10 @@ def test_rendered_attributes(self):
         ibx = get_meta_infobox('Template:Infobox person')
         self.assertEqual(ibx.attributes['native-name'], u'Native name')
 
+    def test_alternate_rendered_attributes_template(self):
+        builder = MetaInfoboxBuilder('Template:Infobox aircraft type')
+        self.assertTrue(builder.rendered_attributes())
```
alvaromorales: This test needs a comment explaining the alternate format (maybe something about `<td>` and `<th>`). You should also point to a sample article, https://en.wikipedia.org/wiki/General_Dynamics_F-16_Fighting_Falcon

The assertion should be more specific; maybe test that "primary-user" maps to "Primary users" (this checks that the infobox template is parsed successfully, and that nbsp (`\xa0`) is removed.

wikipediabase/infobox.py
```diff
@@ -759,31 +759,46 @@ def unescape_lists(val):
 
         tpairs = []
 
-        for row in soup.findall('.//tr'):
-            try:
-                e_key, e_val = row.findall('./*')[:2]
-            except ValueError:
-                continue
+        elems = soup.findall('.//tr')
```
alvaromorales: I find xpath expressions really hard to read. Could you add comments for each one? Some English-readable note like "looks for all tr elements under tag with class foo"

wikipediabase/infobox.py
```diff
@@ -759,31 +759,46 @@ def unescape_lists(val):
...
 
-                val = escape_lists(tostring(e_val))
-                # Extract text
-                val = fromstring(val)
-                val = totext(val)
+                    # replace multiple spaces (including nbsp) with a single space
+                    key = re.sub(r"\s+", u" ", key, flags=re.U).strip()
 
-                # replace nbsp (unicode code 160) with space
-                val = val.replace(unichr(160), u' ')
```
alvaromorales: you need to add these lines back in somewhere, to replace nbsp with normal space (this will get rid of `\xa0`).

wikipediabase/infobox.py
```diff
@@ -759,31 +759,46 @@ def unescape_lists(val):
...
+                    val = escape_lists(tostring(e_val))
+                    # Extract text
+                    val = fromstring(val)
+                    val = totext(val)
 
-                val = unescape_lists(val.strip())
```
alvaromorales: why did you remove the calls to `escape_list` and `unescape_lists`?

wikipediabase/infobox.py
```diff
@@ -759,31 +759,46 @@ def unescape_lists(val):
...
-                tpairs.append((key, val))
+                    # replace nbsp (unicode code 160) with space
+                    val = val.replace(unichr(160), u' ')
+
+                    val = unescape_lists(val.strip())
+                    tpairs.append((key, val))
+        else:
```
alvaromorales: is this `else` for parsing the alternate format from infobox aircraft type? If so, please add a comment and a link to a sample article

alvaromorales:
This is looking great! I added some inline comments. Where possible, try to add explanatory comments. We haven't done a good job in the past on adding comments to the infobox parsing code. This function is one of the hardest in the code base, so comments would be very helpful for the future!

I checked out the PR locally, and tried parsing several articles using the 'Random article' button. I haven't found a single case where this PR introduces a regression. I'm really looking forward for this change to be merged.

wikipediabase/infobox.py
cordeliaavery: @alvaromorales it's still in there in the original code; since i'm parsing using text, not html I wasn't sure that I needed it in alternate format. What are these functions doing, exactly? It's not clear to me why they're here from the regex

alvaromorales:
lgtm! :rocket:

PR #139, opened 2016-01-14
Also remove nbsp in values

alvaromorales:
Added unit tests to check for stray nbsp and bullet point characters [hotfix, master is broken]

cc @cordeliaavery


chsamlee:
Looks good.

PR #149, opened 2016-01-19
Escape super/subscripts in Infobox attributes for formatting purposes

cordeliaavery:
Fixes #105


alvaromorales:
The sort-symbols test for 'Cake' sometimes succeeds locally but fails on Travis. If you revert your changes to `tests/test_sort_symbols.py` Travis tests should pass.

alvaromorales:
Looks good. In `wikipediabase/util.py`, `escape_lists` should be renamed to `escape_tags`, and `unescape_lists` should be renamed to `unescape_tags`. Can you add also add a comment explaining why we whitelist these tags?

alvaromorales:
lgtm! :rocket:

PR #151, opened 2016-01-21
Decouple resolvers and knowledgebase from lispify

### LINKED ISSUE #120: Decouple KnowledgeBase from lispify
Methods in KnowledgeBase return lispified answers.

Let's say we want to connect to a WikipediaBase server using Python. We'd have to un-lispify answers to use in Python. It would be better if we could call methods with RPC and get back native Python data types.

We should move the `lispify` step into TelnetFrontend, and create RPCFrontend using jsonrpc.

### LINKED ISSUE #143: Deprecate typed lists over plists
Per #120, I think it's a good idea to push the lispify logic down to the frontend modules, not the knowledge base or resolvers. This gets complicated because we have to return lists with typecodes.

Python doesn't natively support lists with typecodes. We could wrap the native list with a custom class, but I'd rather look for a simpler solution.

The lispify module does a great job at translating dictionaries to plists. What if we deprecate typed lists over plists? In the external API, this would only involve one change.

Instead of returning:

```lisp
(get "wikibase-term" "Black Sea" (:code "COORDINATES"))
((:coordinates 44 35))
```

we would return:

```lisp
((:coordinates (44 35)))
```

@sfelshin would this be okay?

chsamlee:
Resolves #120 and resolves #143.

External API changes:
- get with coordinates: ((:coordinates 44 35)) -> (:coordinates (44 35))
- get with image-data: ((0 "filename" "some caption")) -> ((:file "filename" :caption "some caption"))

Other important changes:
- LispDate no longer parses date; resolvers are responsible for returning the date with correct format
- multi-key date dictionary such as {"yyyymmdd": "20160121", "html": Jan 21, 2016"} is now supported


tests/test_frontend.py
```diff
@@ -29,8 +29,9 @@ def setUp(self):
         self.fe = Frontend()
 
     def test_simple(self):
+        # extra quotation marks because lispify is moved to frontend
```
alvaromorales: I'm not sure I understand this comment -- what do you mean?

wikipediabase/knowledgebase.py
```diff
@@ -25,22 +24,25 @@ def get(self, cls, symbol, attrs, options={}):
         :param symbol: the Wikipedia article
         :param attrs: the required attributes to get
         :param options: the optional attributes
-        :returns: the attribute's value or an error, lispified
+        :returns: the attribute's value(s) or an error, as a list of dict
         """
-        return lispify([resolve(cls, symbol, attrs, **options)])
+        ret = resolve(cls, symbol, attrs, **options)
+        if not isinstance(ret, list):
```
alvaromorales: Why not have resolvers return a list of dictionaries, instead of checking `isinstance`?

wikipediabase/lispify.py
```diff
@@ -56,13 +55,11 @@ class LispType(Logging):
     priority = 0
     literal = False
 
-    def __init__(self, val, typecode, infobox_attr=None):
+    def __init__(self, val):
         """
-        Lispify a piece of data. Throws LispError on failure.
+        Lispify a piece of data. Throws error on failure.
```
alvaromorales: describe which exception is raised by this method

wikipediabase/resolvers/section.py
```diff
@@ -16,9 +15,9 @@ def _should_resolve(self, cls):
     @provide(name="sections")
     def resolve_section(self, cls, symbol, attr, **options):
         # TODO : implement sections resolver
-        return lispify([])
+        return []
```
alvaromorales: `SectionResolver` returns a list, but `PersonResolver` returns a dictionary (not a list of dictionaries). Why is this?

wikipediabase/resolvers/util.py
```diff
@@ -0,0 +1,38 @@
+import overlay_parse
+
+
```
alvaromorales: +1 on modularity in this PR, it's really great. However, I also think these functions would be useful to classifiers and various other parts of the code. We should move them into `util.py`

wikipediabase/resolvers/util.py
```diff
@@ -0,0 +1,38 @@
...
+    # convert into YYYYMMDD
+    try:
+        d, m, y = date
+        return "%s%04d%02d%02d" % ("-" if y < 0 else "", abs(y), m, d)
+    except:
+        # some error in parsing
+        return None
+
+
+def ret_error(name, **kwargs):
```
alvaromorales: Can we change the name to something more descriptive? Some suggestions are `raise_error` or `return_error`

alvaromorales:
Great work! This is a really big and important change. I'm impressed by your clean solution. I left some inline comments.

I think we should standardize the return value of all resolvers -- have them return lists, instead of checking whether the return value is a list in KnowledgeBase.

tests/test_frontend.py
chsamlee: In the past, frontend itself doesn't call lispify and returns normal string. After the change, the call returns a LispString instance which has an extra pair of quotation marks when converted to string.

alvaromorales:
LGTM, great work! :rocket:

chsamlee:
Changed TermResolver.coordinate's return type when it fails to find the coordinate to None. Returning an empty list won't generate an error.

alvaromorales:
lgtm, glad this was caught before merging :ship:

PR #152, opened 2016-01-21
remove markup-checking lines from lispify

chsamlee:
removed 2 lines from lispify.LispString that removes references and links
these lines should go in resolvers


alvaromorales:
lgtm! :rocket:

PR #157, opened 2016-01-22
Remove punctuation from rendered attributes

### LINKED ISSUE #128: Remove punctuation from rendered attributes
Attributes below should not end in colons.

```python
In [12]: get_meta_infobox('Template:Infobox road').attributes
Out[12]:
{u'allocation': u'Component highways:',
 u'cities': u'Major cities:',
 u'counties': u'Counties:',
 u'decommissioned': u'Existed:',
 u'deleted': u'Existed:',
 u'districts': u'Districts:',
 u'established': u'Existed:',
 u'formed': u'Existed:',
 u'history': u'History:',
 u'junction': u'',
 u'municipalities': u'Municipalities:',
 u'related': u'Related routes:',
 u'restrictions': u'Restrictions:',
 u'significance': u'Cultural significance:',
 u'terminus-a': u'From:',
 u'terminus-b': u'To:',
 u'time-period': u'Time period:',
 u'towns': u'Towns:',
 u'villages': u'Villages:'}
```

chsamlee:
Fixes #128


wikipediabase/util.py
```diff
@@ -242,7 +242,7 @@ def clean_key(e_key):
     # -- expects attribute 'Family:' not 'Family')
     # Will probably want to update this at some point, and can include the following filter:
```
alvaromorales: remove these comments since they are no longer relevant:

```python
    # This formatting can be updated later--currently some test cases expect there to
    # be a colon after the attribute name (see https://en.wikipedia.org/wiki/Template:Taxobox
    # -- expects attribute 'Family:' not 'Family')
    # Will probably want to update this at some point, and can include the following filter:
```

Issue #100, opened 2015-12-03
implement wikibase-term IMAGE-LIVE attribute natively in WikipediaBase


sfelshin:
The `wikibase-term IMAGE-LIVE` attribute is implemented as an Omnibase script purely for hysterical raisins.  It no longer makes a lick of sense to keep this functionality in Omnibase.  We should move it to WikiBase, but note that the Omnibase script allows a file size specification, and we'd have to figure out how to represent that in WikiBase.  Maybe allow additional keyword-value pairs in attribute descriptors (that is, in addition to `:code` and `:rendered`)?  Like this?

    (get "wikibase-term" "Tiger" (:code "IMAGE-LIVE" :size 50))

alvaromorales:
IMAGE-LIVE should return a list of one item containing the first image in the IMAGE-DATA list, optionally resized if `:size` is specified.

E.g.: `(get "wikibase-term" "Bill Clinton" (:code "IMAGE-LIVE" :size 350))` should return the image in Bill Clinton's infobox, resized to 350px.

Wikipedia has a really neat way of resizing images on the server.

Original image URL: https://upload.wikimedia.org/wikipedia/commons/4/49/44_Bill_Clinton_3x4.jpg

URL of same image resized to 350px: https://upload.wikimedia.org/wikipedia/commons/thumb/4/49/44_Bill_Clinton_3x4.jpg/350px-44_Bill_Clinton_3x4.jpg

URL of same image resized to 10px: https://upload.wikimedia.org/wikipedia/commons/thumb/4/49/44_Bill_Clinton_3x4.jpg/10px-44_Bill_Clinton_3x4.jpg

If a size is specified, IMAGE-LIVE needs to make these changes to the image URL:

https://upload.wikimedia.org/wikipedia/commons **/thumb** /4/49/44_Bill_Clinton_3x4.jpg **/SIZEpx-44_Bill_Clinton_3x4.jpg**

Note that an image URL might already contain '/thumb', in which case we only need to append to the end of the URL.

Also, I think this **only** works for images in the Wikipedia Commons ('/commons/' in URL). Here's an example of an image that's _not_ in the commons: https://upload.wikimedia.org/wikipedia/en/8/89/Hanmerst.jpg. I can't get resizing to work. We should investigate this further.

sherryy:
[...]

PR #167, opened 2016-01-27
Implement image-live. Write tests for image-live

sherryy:
implement wikibase-term IMAGE-LIVE attribute natively in WikipediaBase

Fixes: #100


wikipediabase/resolvers/term.py
```diff
@@ -51,6 +52,28 @@ def coordinates(self, article, _, **options):
...
+            if href:
+                if img in href and "File" not in href:
+                    img_url = href
+                    break
+        if "size" in options:
+            size = options["size"]
+            insert_to = re.search("/\w/\w\w/", img_url).start()
+            img_url = img_url[:insert_to] + "/thumb" + img_url[insert_to:] + "/" + str(size) + "px-" + img
+        best["file"] = "https:" + img_url
+        return best
```
chsamlee: You should always return a list of dicts.

chsamlee:
LGTM.

wikipediabase/resolvers/term.py
```diff
@@ -51,6 +52,28 @@ def coordinates(self, article, _, **options):
...
+        if not imgs:
+            return []
+        best = imgs[0]
+        img = best["file"]
+        file_page = get_fetcher().html_source("File:" + img)
+        soup = BeautifulSoup(file_page, 'html.parser')
+        for hyperlink in soup.find_all('a'):
+            href = hyperlink.get('href')
+            if href:
+                if img in href and "File" not in href:
```
alvaromorales: what happens when img is HTML escaped?  For example:

```python
self.term_resolver.image_data("Thoughts of Love", "image-data")
```

returns `"TOL%28Custom%29.jpg"`. Will you find `"TOL%28Custom%29.jpg"` in `href`, or do you need to unescape?

Before line 70 you might need to check if img_url is `None` to be safe?

wikipediabase/resolvers/term.py
```diff
@@ -51,6 +52,28 @@ def coordinates(self, article, _, **options):
...
+                    img_url = href
+                    break
+        if "size" in options:
```
alvaromorales: @chsamlee do you lowercase the keys of `options`? What if someone calls `(get "wikibase-term" "Foo" (:code "IMAGE-LIVE") (:SIZE 100))`?

chsamlee: I believe so (frontend.py, line 47)

alvaromorales:
lgtm :ship:

PR #168, opened 2016-01-28
Get returns a list of HTML values

### LINKED ISSUE #84: Always get infobox values from HTML

### LINKED ISSUE #97: markup appears in :html value (F-16 unit cost)
(get "wikipedia-aircraft-type" "General Dynamics F-16 Fighting Falcon" (:CODE "UNIT-COST"))
    ((:html "F-16A/B: {{US$|link=yes|14.6 million}} (1998 dollars)<ref name=AF_fact_sh/> <br>F-16C/D: {{US$|18.8 million}} (1998 dollars)<ref name=AF_fact_sh/>"))

### LINKED ISSUE #154: Ensure that resolvers remove wikimarkup from answers
These lines were removed from the lispify module:

```diff
         return super(LispString, self).typecode_str()
  
      def val_str(self):
 -        v = re.sub(r"\[\d*\]", "", self.val)  # remove references, e.g. [1]
 -        v = re.sub(r"[[\]]", "", v)  # remove wikimarkup links, e.g. [[Ruby]]
 -        v = v.replace('"', '\\"')  # escape double quotes
 +        v = self.val.replace('"', '\\"')  # escape double quotes
          v = u'"{0}"'.format(v)
          return v
```

We need to ensure that resolvers return clean answers with no wikimarkup.

chsamlee:
"Get" will always return a list of values. When used with "rendered", the function always returns HTML values but markup can be returned when "code" is used for attribute. You can specify the behavior when dealing with markup by passing the "markup" option. The default option is to exclude markup values completely. Other possible options are "allow" (get will return markup values as fallback but always prioritize HTML) and "render" (render markup values into HTML, may impact performance). Examples are given below:
```lisp
(get "wikipedia-aircraft-type" "F-16" (:code "unit-cost"))

(get "wikipedia-aircraft-type" "F-16" (:code "unit-cost") (:markup "allow"))
((:markup "F-16A/B: {{US$|link=yes|14.6 million}} (1998 dollars) <br>F-16C/D: {{US$|18.8 million}} (1998 dollars)"))

(get "wikipedia-aircraft-type" "F-16" (:code "unit-cost") (:markup "render"))
((:html "F-16A/B: US$14.6 million (1998 dollars)<br /> F-16C/D: US$18.8 million (1998 dollars)"))
```
Fixes #84 and fixes #97 and fixes #154.


tests/test_infobox.py
```diff
@@ -64,11 +64,33 @@ def test_exclude_illegal_attributes(self):
...
+    def test_get_code(self):
+        # attributes with rendered name
         ibox = get_rendered_infoboxes("The Rolling Stones")[0]
-        self.assertEqual(ibox.get("origin"), "London, England")
-        self.assertIn("Mick Jagger", ibox.get("current-members"))
+        self.assertEqual(ibox.get("alias", "code"),
+                         [{"html": "The Stones"}])
+        self.assertIn("Mick Jagger",
+                      ibox.get("current-members", "code")[0]["html"])
+        # attributes without rendered name
```
sherryy: Can this one be a separate testcase, with name like test_get_code_attributes_no_rendered_name?

sherryy:
Looks good to me after separate that test case.  In util.py you have " # there is probably more that can be done but is that work necessary?", are you trying to ask Alvaro about this? Otherwise I would delete the comment?

alvaromorales:
This functionality is really nice, great work!

I think the `:markup` option should have three options: `allow`, `render`, `ignore`. The default option should be `render` (I don't think the performance hit will be significant). So if no `:markup` option is passed, the value of `markup` should be `render`.

If `:markup` is set to `ignore`, we should return a descriptive error message, not `attribute-value-not-found`. Perhaps something like `answer-contains-markup`.

@sfelshin, here's the specification of the `get` method and a sample interaction:

> get() should return a list of values. If a :rendered attribute matches multiple :code attributes, all matching attributes should be returned.

```lisp
;; both :code and :rendered
(get "wikipedia-president" "Bill Clinton" (:code "successor" :rendered "Succeeded by"))
((:html "George W. Bush"))

;; both :code and :rendered
(get "wikipedia-president" "Bill Clinton" (:code "successor1" :rendered "Succeeded by"))
((:html "Jim Guy Tucker"))

;; only code
(get "wikipedia-president" "Bill Clinton" (:code "successor"))
((:html "George W. Bush"))

;; only :rendered
(get "wikipedia-president" "Bill Clinton" (:rendered "Succeeded by"))
((:html "George W. Bush") (:html "Jim Guy Tucker") (:html "Frank D. White") (:html "Steve Clark"))
```

sfelshin:
> I think the `:markup` option should have three options: `allow`, `render`, `ignore`.

From a lisp standpoint, it would be safer to have both option names and option values be keywords, e.g., `:allow`, `:render`, `:ignore`.

> If `:markup` is set to `ignore`, we should return a descriptive error message, not `attribute-value-not-found`. Perhaps something like `answer-contains-markup`.

The name `ignore` seems confusing to me.  I guess you intend the option value to mean "if `:markup` is set to `ignore`, then `get` really ought to ignore any values containing markup, so if markup appears in any values, that's an error".  But it also sounds like it could mean "if any values contain markup, ignore that problem and return, er, it's not clear what".  Can we come up with a less ambiguous name?  What do you think about `error` (or rather, `:error`)?

Also, can we have a test case for multiple values where some are valid values and some improperly contain markup?  You'd have to either find or fake a class–article–attribute combination with mixed html and markup answers.  E.g., something like along these lines:

```
(get "wikipedia-president" "Bill Clinton" (:rendered "Succeeded by") (:markup :error))
((:html "George W. Bush")
 (:html "Jim Guy Tucker")
 (:error answer-contains-markup "Frank D. White: {{US$|link=yes|14.6 million}} (1998 dollars)")
 (:html "Steve Clark"))
```

alvaromorales:
Because Python doesn't support keywords natively, It's hard to have option values be keywords. Now that we're using `lispify` in other projects (e.g. to generate annotations), changing the implementation would cause breaking changes. Perhaps you can enforce validation on the START side?

I can see why `ignore` is confusing. `error` is confusing in terms of the WikipediaBase implementation, because we don't throw an error if an answer contains markup, we just skip it. If we ultimately fail to find any answers for the query, we then throw the error.

I think `disallow` makes more sense (conveniently, the antonym of `allow`). `:markup` is then one of:

* `render`: render markup values as HTML
* `allow`: return markup value unparsed
* `disallow`: don't return answers that contain markup

sfelshin:
> Because Python doesn't support keywords natively, It's hard to have option values be keywords. ... Perhaps you can enforce validation on the START side?

Okay.

> I think `disallow` makes more sense (conveniently, the antonym of `allow`).

`disallow` still seems ambiguous.  I admit it's convenient that it's the antonym of `allow`, but given that you just described this option as "if an answer contains markup, we just skip it", how about `skip`?

sfelshin:
Say, have we decided whether it's possible for `get` to return multiple errors or errors intermingled with valid values, as opposed to only a unitary list error `((:error ...))`?  Based on what you say above, it sounds like we only support a unitary list error.  On the one hand, that seems convenient, but on the other hand, it suppresses information.  If we got a return value like

```
(get "wikipedia-president" "Bill Clinton" (:rendered "Succeeded by") (:markup :error))
((:html "George W. Bush")
 (:html "Jim Guy Tucker")
 (:error answer-contains-markup "Frank D. White: {{US$|link=yes|14.6 million}} (1998 dollars)")
 (:html "Steve Clark"))
```

that would let the caller know whether or not the answer was complete.  We could answer, e.g., "Bill Clinton was succeeded by George W. Bush, Jim Guy Tucker, Steve Clark, and additional values that cannot be displayed."

chsamlee:
The lispify module is capable of handling valid values mixed with errors, but we do not have mixed types under the current implementation:
- if `get` is called with `(:rendered attr)`, then we only look at HTML and the output is always a list of valid answers
- if `get` is called with `(:code attr)`, then the list has at most one item. This is because the attributes' `code` is unique within the infobox.

sfelshin:
Okay, works for me.

chsamlee:
Some statistics on `get`'s speed:
(1000 runs of '(get "wikipedia-aircraft-type" "F-16" (:code "unit-cost"))') with each markup option)
skip: 2.74347496033 ms/call
allow: 2.57662796974 ms/call
render: 3.12301492691 ms/call

wikipediabase/infobox.py
```diff
@@ -135,24 +137,99 @@ def __init__(self, symbol, template, cls, markup, html):
...
         self.html = html
 
-    def get(self, attr, source=None):
+    def get(self, attr, attr_type, markup="render"):
         """
-        - First try plain markup attributes
-        - Then translating each markup's translations
+        :param attr: attribute name
+        :param attr_type: attribute type, which is either "code" or "rendered"
+        :param markup: "disallow": skip markup value
```
alvaromorales: change `"disallow"` to `"skip"`

alvaromorales:
lgtm :ship:  (you need to fix merge conflicts)

PR #169, opened 2016-01-29
Fixing TermResolver's coordinates query

### LINKED ISSUE #150: wikibase-term (:code "COORDINATES") --> (:attribute-value-not-found ...)
```
(get-attributes "wikibase-term" "Statue of Liberty")
((:code "SHORT-ARTICLE") (:code "URL") (:code "NUMBER")
 (:code "COORDINATES") (:code "WORD-COUNT")
 (:code "IMAGE-DATA") (:code "PROPER"))
(get "wikibase-term" "Statue of Liberty" (:code "COORDINATES"))
```

but when I look at https://en.wikipedia.org/wiki/Statue_of_Liberty I can clearly see coordinates at the upper right hand side of the page (right-flushed opposite "From Wikipedia, the free encyclopedia").

chsamlee:
Fixes #150


tests/test_resolvers.py
```diff
@@ -242,6 +242,15 @@ def test_special_attributes_coordinates(self):
                          '((:coordinates (10.5 -66.9167)))')
         # XXX: according to wikipedia not so
 
+    def test_special_attributes_coordinates_not_in_infobox(self):
+        coord = resolve('wikibase-term', 'Hong Kong', {'code': 'coordinates'})
+        self.assertTrue([{'coordinates': [22.3, 114.2]}])
```
alvaromorales: You should probably call `resolve('wikibase-term', 'Hong Kong', {'code': 'coordinates'})` a second time and repeat the assertion. We need a test for the strange only-works-the-first-time error (but please add a code comment to explain why we need to call the resolver twice).

alvaromorales:
lgtm :ship:

Issue #156, opened 2016-01-22
Change TelnetServer to support multithreading


alvaromorales:
TelnetServer is currently single-threaded. This can be fixed by adding `ThreadingMixIn`. However, some work needs to be done before:

* Lock Fetcher and Renderer when fetching from wikipedia.org, not the DB. This need to be done to play nice with Wikipedia's API [etiquette](https://www.mediawiki.org/wiki/API:Etiquette).
* Make LRUCache thread-safe
* Ensure resolvers and classifiers are stateless

alvaromorales:
I just noticed that the TelnetServer does not close long connections via a timeout. A single-threaded server is dangerous, especially if there is no timeout. I've marked this as a bug.

Here's how I reproduced the issue:

Window 1: connect but don't send a command (keep the connection open)
```
$ telnet wikibase-host 8023
Connected to wikibase-host
Escape character is '^]'.
... [no input] ...
```

Window 2: connect and send a command

```
$ telnet wikibase-host 8023
Connected to wikibase-host
Escape character is '^]'.
(get-classes "Bill Clinton")
... [no output] ...
```

The TelnetServer accepts the request, but blocks until Window 1 sends a command or closes the connection. Window 1's connection never times out.

PR #171, opened 2016-01-30
Allow newlines as whitespace in WikipediaBase requests

### LINKED ISSUE #170: Allow newlines as whitespace in WikipediaBase requests
WikipediaBase doesn't allow newlines in requests, e.g., you can send
```
(get "wikipedia-terrorist-attack" "Centennial Olympic Park bombing" (:CODE "DATE"))
```
but not
```
(get "wikipedia-terrorist-attack" "Centennial Olympic Park bombing"
(:CODE "DATE"))
```
or
```
(get "wikipedia-terrorist-attack" "Centennial Olympic Park bombing" (:CODE
 "DATE"))
```

If requests are proper Lisp forms, then spaces, newlines, tabs, and sequences of them are all valid whitespace.  When I test, I find that I can use spaces, tabs, and sequences of them as whitespace in requests, but not newlines.

This is not very important, so if it's a pain to handle newlines in requests in python, say so and close this issue. Otherwise it can hang around as a low-priority enhancement until we run out of more interesting issues.

chsamlee:
Fixes #170


sfelshin:
It looks to me like the fix is to replace all sequences of any kind of whitespace in a request with a single tab.  As far as I know, symbol and attribute names don't contain tabs or multiple spaces (or newlines!), but synonyms can contain multiple spaces, or so I find by checking navassa:/scratch/wikipedia/wiki-2015-12-01/wikibase-term-generated-synonyms.  (All whitespace in synonyms is reduced to a single space when symbols and synonyms are loaded into Omnibase, but that's a loss of information we'd like to stop someday.)  Also, the ~synonyms that are the first argument passed to the `sort-symbols-named` command are whatever the user typed and can therefore contain funky whitespace.  And unless there is a Wikipedia stricture against tabs or multiple spaces in symbol and attribute names, we should allow for it.  How about adding test cases for:
   * symbols and attributes—maybe you can find a real case, but here's a fake:

<pre>
(get "wikipedia-something" "Multiple    spaces" (:code "MULTIPLE     SPACES"))
</pre>

   * synonyms—here's a real example (with some symbol names pruned out for brevity):

<pre>
(sort-symbols-named "black    sea" "Crytek Black Sea" "Black Sea" "Black Sea (Christian Fennesz album)" "Black Sea Region" "Black Sea (XTC album)")
</pre>

chsamlee:
I can change the code so that whitespace characters are replaced with space 1-to-1 if the current system can already handle multiple spaces (the one with ASCII code 32). The more complex problem is to change the telnet server to take multiple lines: if we see a query like `(get-classes "F-16"`, we have no idea if the user intends to input a multi-line query or simply forgets the ending parentheses.

alvaromorales:
Once the telnet server supports multithreading (#156), we can just time out on incomplete queries like `(get-classes "F-16"`. We can't do this currently because the server would block during the timeout period.

On the other hand, how do we detect valid queries like `(get-classes "F-16")`? Do we need to keep track of balanced parentheses?

chsamlee:
Parentheses tracking seems to be the easiest method to me as we can simply count left parentheses and right parentheses.

sfelshin:
> Parentheses tracking seems to be the easiest method to me as we can simply count left parentheses and right parentheses.

Yes, although please be careful not to count parentheses inside strings.

alvaromorales:
With #177 merged, this PR is now unblocked and can be continued

chsamlee:
(only the last commit contains the actual changes)

alvaromorales:
Very cool! A couple of things:

```
$ telnet host 8023
(get-classes "Bill Clinton"))))))
(SyntaxError(LexToken(LIST_END,u')',1,28),))
```

This is **not** a properly escaped error. The return format should be something like `((:error SyntaxError :message "LexToken(LIST_END,u')',1,28)"))`. It looks like the error handling code is not deadling with this case properly. Can you fix it and add a test case?

Also, for this request:

```
$ telnet host 8023
(
get
"wikibase-term"
"Bill Clinton"
(:code "NUMBER")
)
```

The server log prints out the following:

```
... Received request: (get"wikibase-term""Bill Clinton"(:code "NUMBER"))
```

You'll notice that the arguments of the request are not separated by spaces (the log should print `(get "wikibase-term" "Bill Clinton" (:code "NUMBER")`). Can we perhaps add a space when the input is a newline character?

sfelshin:
FYI, lisp will have no trouble reading
```
(get"wikibase-term""Bill Clinton"(:code "NUMBER"))
```
despite the lack of whitespace separators.  However, it would be tasteful, and helpful to human readers, to separate elements with whitespace.

chsamlee:
Reason for having 2 levels of try-except in frontend.py: when `loads(string) (line 67) throws an exception, the message instance is a custom class instead of a string which should be the case in every sane program.

alvaromorales:
lgtm 🚀

PR #173, opened 2016-02-02
Making empty options plist work

### LINKED ISSUE #172: Empty options plist causes IndexError
The `get` API is: `(get class symbol attribute-plist options-plist)`.

`options-plist` is an _optional_ argument, so the caller can specify no options or pass an empty list.

Passing an empty plist of options causes an error:

```lisp
(get "wikipedia-sea" "Black Sea" (:code "AREA") ())
((:error IndexError :message "list index out of range"))
```

chsamlee:
Fixes #172


wikipediabase/frontend.py
```diff
@@ -55,7 +59,11 @@ def _eval(self, ls):
 
     def eval(self, string):
         ls = loads(string)
-        ans = self._eval(ls)
+        if len(ls) > 0:
+            ans = self._eval(ls)
+        else:
+            # empty query '()'
+            ans = [return_error("BadRequestFormat", message="empty query")]
```
alvaromorales: add a test case for this

alvaromorales:
lgtm :rocket:

chsamlee:
Something to note: should we treat the empty query `()` as a bad request?

alvaromorales:
Yes.

I like that the error you're using is more descriptive, we were previously throwing `((:error IndexError :message "list index out of range"))`

PR #175, opened 2016-02-03
Improve wikibase-person classifier regexes

### LINKED ISSUE #174: Plato should be a wikibase-person
In revision 1.0.0 on navassa (currently the START production host):
```
$ telnet navassa 8023
(get-classes "Plato")
("wikibase-term" "wikibase-sections" "wikipedia-philosopher" "wikibase-person")
```
In revision 1.1.0 running experimentally on kiribati:
```
$ telnet kiribati 8023
(get-classes "Plato")
("wikibase-term" "wikibase-sections" "wikipedia-philosopher")
```

Why did Plato lose his humanity?

chsamlee:
Fixes #174


wikipediabase/classifiers.py
```diff
@@ -65,8 +65,8 @@ class PersonClassifier(BaseClassifier):
 
     CATEGORY_REGEXES = [re.compile(p) for p in [
         ".* person",
-        "^\d+ deaths.*",
-        "^\d+ births.*",
+        "^\d+((st|nd|th)-century|0s)?( bc)? deaths.*",
+        "^\d+((st|nd|th)-century|0s)?( bc)? births.*",
```
alvaromorales: `rd` is probably missing here, as in "3rd century"

alvaromorales:
`(get-classes "Aristotle")` fails

alvaromorales:
It seems like Greek philosophers have ganged up against us!

Failures:
* Socrates
* Pythagoras
* Thales
* Democritus
* Epicurus
* Diogenes of Sinope
* Leucippus
* Zeno of Elea
* Crates of Thebes

chsamlee:
This is interesting. Only Crates of Thebes is failing for me, and it is because the article uses "year of birth unknown" category instead of the typical "year of birth missing".

alvaromorales:
Alas, I spoke too soon. I was running a different branch of the code, sorry about that. I've edited my comments above.

Only a few Greek philosophers are causing trouble. And some Chinese philosophers as well: Wang Fu, Lie Yukou, Yang Xiong, Zhang Sanfeng, Zhuangzi, Han Feizi, Li Kui, Yang Zhu, Hui Shi, Xu Xing, Su Qin, Zhang Yi, Yue Yi, Shan Tao, Xiang Xiu.

More here: https://en.wikipedia.org/wiki/List_of_Chinese_philosophers

For these cases, we should distinguish between badly-written articles (we can just edit those on Wikipedia) and valid patterns we're omitting.

alvaromorales:
Everything in Wikipedia is hard! Some of the Chinese philosopher test cases were failing because they point to disambiguation pages. I updated the list.

alvaromorales:
If we have these patterns:

```python
        ".* goddesses",
        ".* musicians",
        ".* players",
        ".* singers",
```

It seems equally as arbitrary to add `".* philosophers"`.

But maybe you can come up with a more robust, language-based strategy. We could get a list of nationalities and occupations from START, but I'd like to avoid lexicon data files if possible.

chsamlee:
I only recognize half of these people, and even fewer if I can only rely on pinyin. Eastern Asian languages have lots of homophones. [Here](https://en.wikipedia.org/wiki/One-syllable_article)'s a potentially interesting read.

The main problem is that many non-western philosophers with unknown birth date is not under the category of "birth date missing/unknown". In cases like Su Qin, nationality is the only clue we have. An option is to use suffixes, but there are many edge cases (Greek) and cases that would overkill (-i for Iraqi, Israeli). On the other hand, directly reading from a list will degrade performance significantly.

What were some false positives we have with ".* person"?

alvaromorales:
See the discussion in #116 for issues with .* person

Here's a list of nationalities we've used in the past: https://raw.githubusercontent.com/infolab-csail/whoami/5e48ad7ac2966642f6fb12f25d45c221b7a23c1e/src/main/resources/nationalities.txt

alvaromorales:
I think we should merge this PR and open issues for more difficult enhancements to the way we classify people.

wikipediabase/classifiers.py
```diff
@@ -63,10 +63,14 @@ def classify(self, symbol):
...
+    _DATE_REGEX = "^\d+((st|nd|rd|th)-century|0s)?( bc)?"
+
     CATEGORY_REGEXES = [re.compile(p) for p in [
         ".* person",
-        "^\d+ deaths.*",
-        "^\d+ births.*",
+        _DATE_REGEX + " deaths.*",
+        _DATE_REGEX + " births.*",
+        # (date) (nationality) (people), but misses things like Anglo-Irish
+        _DATE_REGEX + " [a-z]+ people",
```
alvaromorales: you could change it to `[a-z-]` to capture Anglo-Irish

alvaromorales:
The `*person` regex is causing false positives, see "Armoured personnel carrier", "IBM Personal System/2"

I started a job that loops through every article in the backend and prints out those that have the "wikibase-person" class. See nauru:/scratch/alvarom/pywikipediabase/people. It's a good way to find false-positives.

sfelshin:
https://en.wikipedia.org/wiki/The_Crystal_Method should not be classified as a person.  It's a duo.

Ditto https://en.wikipedia.org/wiki/QuickTime.  Note non-person categories such as "OS X media players".

Ditto https://en.wikipedia.org/wiki/Six_Feet_Under_%28TV_series%29.  Is it the fault of "actors" in category "Outstanding Performance by an Ensemble in a Drama Series Screen Actors Guild Award winners"?  "Actors" shouldn't count there since it isn't the last word (nor the first noun, in case that's your metric).

Ditto https://en.wikipedia.org/wiki/Eight_Immortals, which is a group of ~people.

chsamlee:
For the QuickTime case, I think it is actually the `.* player` category that leads to misclassification. As for The Crystal Method case, it is because the wikipedia musician categories don't distinguish between individuals and groups. We can add in a check that says if a category contains the word "groups" or "organizations" then it is not a person. This doesn't solve the Eight Immortals problem though.

The *person and *actor regex should only match the end of string. Should be an easy fix.

tests/test_classifiers.py
```diff
@@ -64,15 +72,13 @@ def test_real_animals(self):
     def test_alien_characters(self):
         # Fails because pages in "Fictional Characters" category
         # are currently classified as 'wikibase-person'
+        # Update: only Jabba the Hut fails
```
alvaromorales: Please remove tests that no longer fail (or move them to some other method)

alvaromorales:
looks good!

Issue #77, opened 2015-10-19
Errors parsing birth and death dates from the first sentence


sfelshin:
Richard Shakespeake was born in 1490 so he can't possibly still be alive, yet in Python WikiBase:

(get "wikibase-person" "Richard Shakespeare" (:code "BIRTH-DATE"))
((:YYYYMMDD 14900000))
(get "wikibase-person" "Richard Shakespeare" (:code "DEATH-DATE"))
((:ERROR ATTRIBUTE-VALUE-NOT-FOUND :REPLY "Currently alive"))

Compare Ruby WikiBase:

(get "wikipedia-person" "Richard Shakespeare" (:code "BIRTH-DATE"))
((:YYYYMMDD 14900000))
(get "wikipedia-person" "Richard Shakespeare" (:code "DEATH-DATE"))
((:YYYYMMDD 15610000))

Richard Shakespeare's first sentence says "Richard Shakespeare (1490–d. before 10 February 1561) was a husbandman of [snip]".  The death date is hard to parse, but python WikiBase should something reasonable, which could be:

  - return 1561 like ruby WikiBase
  - return a non-yyyymmdd value: ((:html "d. before 10 February 1561"))
  - deprecated option: return the empty set of values: ()
  - return the value both ways: ((:html "d. before 10 February 1561" :yyyymmdd 15610210))

sfelshin:
I meant to say before: The presence of "–d. " after an apparent birth date should be a sufficient clue that the person is dead, even if the date(?) of death is too hard to interpret.

alvaromorales:
@fakedrake where should the Richard Shakespeare bug be fixed? In WikipediaBase or overlay_parse?

The first sentence is:

> Richard Shakespeare (1490–d. before 10 February 1561) ...

This follows a common pattern: dates in parenthesis in the first sentence; birth date separated from death date by a hyphen. We should ignore periods in the parens.

alvaromorales:
I found this example using Wikipedia's "Random Article".

```
(get "wikibase-person" "Adarnase II of Klarjeti" "BIRTH-DATE")
((:yyyymmdd 09450000))

(get "wikibase-person" "Adarnase II of Klarjeti" "DEATH-DATE")
((:error attribute-value-not-found :reply "Currently alive"))
```

when the first sentence reads:

> Prince Adarnase II (Georgian: ადარნასე) (died 945) was a Georgian royal prince of the Bagrationi dynasty of Tao-Klarjeti branch.

[...]

PR #176, opened 2016-02-03
Don't default to 'currently alive' when a birth date is not found

chsamlee:
See #77 for details.


tests/test_resolvers.py
```diff
@@ -193,6 +193,12 @@ def test_birth_death_date_from_article(self):
                          # '((:html "428/427 or 424/423 BC"))'
                          '((:yyyymmdd -04260000))')
 
+    def test_date_unknown_death_or_birth(self):
+        birth = resolve('wikibase-person', 'Homer', {'code': 'birth-date'})
+        self.assertTrue(isinstance(birth[0], Exception))
+        death = resolve('wikibase-person', 'Homer', {'code': 'death-date'})
+        self.assertTrue(isinstance(death[0], Exception))
```
alvaromorales: You need to check the error messages here, not just that it's an exception. It should be `Unknown` instead of `Currently alive`.

tests/test_resolvers.py
```diff
@@ -244,13 +250,13 @@ def test_special_attributes_coordinates(self):
...
         coord = resolve('wikibase-term', 'Hong Kong', {'code': 'coordinates'})
-        self.assertTrue([{'coordinates': [22.3, 114.2]}])
+        self.assertTrue(coord == [{'coordinates': [22.3, 114.2]}])
         # make sures the only-works-the-first-time error is fixed
         # (in the past calling the same coordinates query the second time
         # gives an error because the HTML tree is accidentally modified; see
         # util.totext for more details)
         coord = resolve('wikibase-term', 'Hong Kong', {'code': 'coordinates'})
-        self.assertTrue([{'coordinates': [22.3, 114.2]}])
+        self.assertTrue(coord == [{'coordinates': [22.3, 114.2]}])
```
alvaromorales: How about `self.assertEqual([{'coordinates': [22.3, 114.2]}], coord)` for consistency with other tests?

alvaromorales:
How about adding a heuristic to `death_date` to improve precision?

> "People don't live more than 150 years"

If the birth year from a parsed `yyyymmdd` date is less than `current_year - 150` (and it's not 0000 for unknown), we raise error `Unknown`. Otherwise, we raise error `Currently alive`.

Of course, we still need to greatly improve our ability to parse dates.

alvaromorales:
What's the status on this?

chsamlee:
Most likely it will be implemented sometime tomorrow.

wikipediabase/resolvers/person.py
```diff
@@ -92,9 +94,14 @@ def death_date(self, symbol, _, **options):
         if death_date:
             return [death_date]
         else:
-            if self.birth_date(symbol, _, **options):
-                return [return_error('AttributeValueNotFound',
-                                     reply='Currently alive')]
+            birth_date = find_date(symbol, 'birth-date', **options)
```
alvaromorales: this heuristic needs a comment, something like:

> "People don't live more than 150 years"
> If the birth year from a parsed yyyymmdd date is less than current_year - 150 (and it's not 0000 for unknown), we raise error Unknown. Otherwise, we raise error Currently alive.

tests/test_resolvers.py
```diff
@@ -193,6 +193,20 @@ def test_birth_death_date_from_article(self):
                          # '((:html "428/427 or 424/423 BC"))'
                          '((:yyyymmdd -04260000))')
 
+    def test_date_unknown_death_or_birth(self):
+        birth = resolve('wikibase-person', 'Homer', {'code': 'birth-date'})
+        self.assertTrue(isinstance(birth[0], Exception))
+        self.assertTrue(type(birth[0]).__name__, "AttributeValueNotFound")
```
alvaromorales: shouldn't it be `assertEquals` instead of `assertTrue`?

alvaromorales:
looks good! :rocket:

PR #177, opened 2016-02-03
Support multi-threading

### LINKED ISSUE #156: Change TelnetServer to support multithreading

alvaromorales:
Made LRUCache, CachingFetcher, and CachingRenderer thread-safe by using locks.

Added a global lock for requests to the Wikipedia API (to comply with the API etiquette). This lock is used by Fetcher and Renderer.

Added ThreadingMixIn to TelnetServer to support concurrent requests. Added socket timeout for long-running client connections.

Cleaned up command-line arguments.

Fixes #156


chsamlee:
Looks good to me.

PR #178, opened 2016-02-03
Fix encoding issues with lispify

### LINKED ISSUE #160: lispify.py: __str__ functions should return string and not unicode
`__str__` functions in `lispify.py` appear to return unicode strings instead of ascii strings. This becomes a problem when one calls, for example, `str(lispify(v))` when `v` is an object that contains a unicode character. Unicode strings should be encoded before returning, or be returned through `__unicode__`.

Note that these `__str__` functions make a call to `util.output` function, which should have been removed for production according to the internal comment.

alvaromorales:
Also removed old assertions and outdated TODO comments.

Fixes #160


moldot:
LGTM! Sorry for the delay. Once merged, I can try running the annotation code to see if the unicode problem is gone.

PR #183, opened 2016-02-19
Handle issue with infoboxes when markup and latest HTML are inconsistent

### LINKED ISSUE #96: Homer's lack of birth and death dates should result in attribute-value-not-found, not AssertionError
(get "wikibase-person" "Homer" (:CODE "BIRTH-DATE"))
    ((:error AssertionError :message ""))

I think the correct output would be

    ((:error attribute-value-not-found ...))

(As far as I know, we currently have no specification for the elliptical material, and START will be okay the value is simply `((:error attribute-value-not-found))`.)

alvaromorales:
When markup in the backend is sometimes inconsistent with the latest HTML from wikipedia.org (when an article is updated). In extreme cases, this inconsistency breaks our ability to extract HTML infoboxes. For example, in late Oct '15 the infobox in Homer's article was deleted. The markup in the backend had a version where the infobox was present. In this case, we can only fall back to using the markup in the backend. We can't use the latest HTML, because the Infobox object built by InfoboxBuilder (used by get-classes and get-attributes) only reads from the backend. We don't want to be in a situation where get-attributes returns a list of attributes that RenderedInfobox does not have. Updating to the latest dump usually fixes this problem.

Also added documentation on the painful process of matching markup infoboxes to HTML infoboxes.

Tested that this change works with the Dec '15 dump and "Susan Pamerleau", where the markup in the backend contained an infobox that has since been deleted from the latest HTML page from wikipedia.org.

Fixes #96


chsamlee:
LGTM.

PR #185, opened 2016-03-03
Fixed article's paragraph classification

### LINKED ISSUE #184: Incorrect classification of article paragraphs
The query
```lisp
(get "wikibase-term" "An Weijiang" (:code "short-article"))
```
gives an error. Upon investigation, it seems to be caused by the failure of wikipediabase.article.Article to generate paragraphs from the Wikipedia article.
```python
from wikipediabase.article import get_article
get_article("An Weijiang").paragraphs()  # empty list, which will lead to an error
```
Also, the paragraphs should not include HIDDEN ERROR messages in the beginning. (This error is non-fatal.)
```python
from wikipediabase.article import get_article
# should be "Julian Berntzen (born ...", but returns "HIDDEN ERROR: ..."
get_article("Julian Berntzen").first_paragraph()
```

chsamlee:
Fixes #184


alvaromorales:
lgtm :rocket:

PR #192, opened 2016-03-15
Add models from wikithingsdb to backend and get rendered attributes from nested infoboxes

### LINKED ISSUE #66: Integrate WikithingsDB into WikipediaBase
WikithingsDB provides a whole bunch of data that could be used by WikipediaBase and START. We need to add prefixes to avoid namespace conflicts, but the general API is:

(get-types “Bill Clinton”)
(get-classes “Bill Clinton”)
(get-hypernyms “Bill Clinton”)
(get-class-hypernyms “wikipedia-president”)
(get-articles-of-type “French mathematician”)
(get-articles-of-class “wikipedia-president”)
(get-hyponyms “company”) ==> airline, bank, brewery, …
(get-synonyms “Bill Clinton”)

Any ideas on "The Right" namespacing?

**EDIT: see my second post that broadens the scope of the discussion of how best to integrate WikithingsDB into WikipediaBase.**

### LINKED ISSUE #155: Mark articles as redirects or disambiguation pages when creating the backend

### LINKED ISSUE #190: Get rendered attributes from nested infoboxes
To map code attributes to rendered attributes (e.g. `SUCCESSOR` --> "Succeeded by"), we look at infobox template documentation pages. These pages show examples of how to use the infobox template, and contain "representative" attributes. For example, the doc page for [infobox person](https://en.wikipedia.org/wiki/Template:Infobox_person/doc) mentions common attributes for people, like `BIRTH_DATE` and `KNOWN_FOR`.

[Adele](https://en.wikipedia.org/wiki/Adele) is classified as `wikipedia-person`. We don't get rendered mappings for attributes like `INSTRUMENT` ("Instruments"), `LABEL` ("Labels"), or `GENRE` ("Genres"). We can parse Adele's infobox just fine -- the problem is that the doc page for `Template:Infobox person` doesn't list attributes like `INSTRUMENT`. My guess is that these attributes aren't common enough across all people, so the editors don't bother to list them.

With the new backend, we precompute classes for all articles, and can easily get all articles with class `wikipedia-person`. Instead of just looking at the template doc page, we can look at each article with class `wikipedia-person` and gather a complete list of code attributes. We can then use the existing MetaInfoboxBuilder infrastructure to get more rendered attributes.

This will inevitably slow down the `generate_meta_infoboxes.py` script. But since this script is run at compile time, it shouldn't be a big problem. We should experiment with the number of articles to sample (perhaps 1000 per class).

`MetaInfoboxBuilder` should support two modes, a 'fast' one (only look at doc subpages) and a 'complete' one (look at a representative sample of articles).

alvaromorales:
Added models from wikithingsdb to the WikipediaBase backend. It makes more sense to have this functionality within WikipediaBase. Types are definitions from the first sentence of articles, extracted by whoami. Classes are infobox classes. Added queries like `articles_of_class` and `types_of_article`. Still need to add hypernyms.

Building the backend now takes significantly longer. Because of this, I've tried to speed it up using Python's `threading` and then `multiprocessing`. I ended up choosing `multiprocessing` because of the simpler logic, but may revert to `threading`. I'll leave this for a future PR, since this is a good stopping point.

Added `get-types` and `get-server-info` methods to the knowledge base.

With `get-types`, we can fetch definitions from the first sentence. Definitions are extracted by whoami (at compile time). Example usage:

```lisp
> (get-types "Victor Kac")
("mathematician" "american mathematician" "american" "soviet mathematician" "soviet")
```

With `get-server-info`, we can see the WikipediaBase server version and the timestamp of the Wikipedia dump used to generate the backend. Example usage:

```lisp
(get-server-info)
((:backend "2016-02-04" :server "1.2.0"))
```

Added a way to get rendered attributes for nested infoboxes. For example, Adele's wikipedia-person infobox has a nested wikipedia-musical-artist infobox with attributes like `GENRE`. We now indicate the source template for each markup attribute, and use this template to look up rendered attributes in the corresponding MetaInfobox.

Updated requirements.txt and added scripts/script-requirements.txt. Scripts require some large dependencies that the WikipediaBase server doesn't need, like spaCy. To speed up build times, we split requirements into two files. Only those who rebuild the backend will need to install script-requirements.txt.

Added scripts/generate_graph.py to get a graph of links between Wikipedia articles. This script works, but is currently too slow (it would take an estimated 100 days to complete). We should transform the script into a Spark job.

(cc @michaelsilver -- fyi, regarding WikithingsDB)

Resolves #66 and fixes #190 and fixes #155


alvaromorales:
@aadah please keep your commits on your own branch, instead of merging them into this branch. When you're ready to merge your changes into master, open a pull request.

moldot:
This is quite a large pull request with mostly code I haven't worked with, so I could just glance over and everything looks fine : ) 

I'm gonna soon push `types_of_class` in `dbutil.py`. It works, but I'm not sure if how I wrote it is the best way. Not sure how `peewee` works.

PR #194, opened 2016-03-26
Fix issue with un-canonicalized infobox templates

alvaromorales:
> In wikimarkup, if a link target begins with a lower case letter, it will be interpreted as if it began with the equivalent capital letter. The link target is case-sensitive except for the first character (so [[atom]] links to "Atom" but [[ATom]] does not).

https://en.wikipedia.org/wiki/Help:Link#Conversion_to_canonical_form

The same applies to infobox templates. The template for `wikipedia-book` may be specified as `infobox book` or `Infobox book`. The latter fails if we don't canonicalize the wikilink.

Fixed it so infobox template links are correctly canonicalized.


moldot:
LGTM!

PR #196, opened 2016-03-29
Add missing comma to list of regexes to prevent infinite loop

alvaromorales:
You heard it right ... one teeny missing comma in a list of regexes causes WikipediaBase to grind to a halt with the query `(get-classes "Clifford P. Case")`.

One of Python's horribly dangerous "features" is to automatically concatenate strings in a list. `["a", "b"]` is a 2-element, list, but `["a" "b"]` (note the missing comma) is evaluated as a 1-element list `["ab"]`. This allows you to break up a string into multiple lines, but it's horribly dangerous! A comma is so easy to miss when you're editing a file or doing a code review. :scream: :scream: :scream:

**Lists should _always_ have trailing commas**.

Commit 7ee9dce added a pattern to `CATEGORY_REGEXES` in `PersonClassifier`, but did not include a trailing comma. This caused two regexes –`"people from(( |, )?[a-z]+)+"` and `".* person"`– to be merged into the pattern `'people from(( |, )?[a-z]+)+.* person'`. Then, poor old [Clifford P. Case](https://en.wikipedia.org/wiki/Clifford_P._Case) comes along, with the category "people from franklin township, somerset county, new jersey". Matching this category to the merged regex causes WikipediaBase to go into an _infinite loop_.

This explains why we've had to restart WikipediaBase in production twice since we deployed v.1.2.0.

Added a test for Clifford P. Case.


alvaromorales:
We're running this unreleased branch in production. I branched off of the `v.1.2.0` tag and committed this hotfix.

Ignore the failing Travis tests -- please review anyway. I'll merge with master before merging this PR.

chsamlee:
LGTM. That missing comma is surely disastrous!

PR #198, opened 2016-03-30
Fixed rendered attributes that do not appear in HTML items

### LINKED ISSUE #195: Why doesn't attribute (:rendered "Assumed office") work?
There are several attributes that render as `Assumed office`:
```
$ telnet kiribati 8023
(get-attributes "wikipedia-president" "Bill Clinton")
(... (:code "TERM-START" :has-value t :rendered "Assumed office")
 (:code "TERM-START1" :has-value t :rendered "Assumed office") ...
 (:code "TERM-START2" :has-value t :rendered "Assumed office")
 (:code "TERM-START3" :has-value t :rendered "Assumed office") ...)
```
But WikipediaBase fails to find any of them:
```
$ telnet kiribati 8023
(get "wikipedia-president" "Bill Clinton" (:rendered "Assumed office"))
```
Note that the attributes can be found by their `:code` names:
```
$ telnet kiribati 8023
(get "wikipedia-president" "Bill Clinton" (:code "TERM-START1"))
((:html "January 11, 1983"))
```
FYI, I noticed this while looking for a test case of an attribute that returns multiple values.

chsamlee:
Fixes #195.


wikipediabase/infobox.py
```diff
@@ -162,18 +162,43 @@ def get(self, attr, attr_type, markup="render"):
...
+        # process a markup answer according to specified option
+        def _process_markup(markup_ans):
+            if is_clean_markup(markup_ans):
+                return {"html": clean_value(markup_ans)}
+            elif markup == "skip":
+                return None
+            elif markup == "allow":
+                return {"markup": markup_ans}
+            elif markup == "render":
+                return {"html": clean_value(get_renderer().render(markup_ans))}
```
alvaromorales: do we need to call `clean_value` on the output of `render()`? is it safe / tested?

chsamlee: The function should be safe since it is tested in :code options before. I also added a test case for Bill Clinton's "assumed office" attribute.
As for `clean_value`: it is probably unnecessary if the markup is clean enough (no references, no nbsp, etc). Some functionalities in `clean_value` are not needed if `get` is always the last query that START sends to wikipediabase, e.g. taking out HTML comments. How good is wikiextractor at cleaning wiki markup?

chsamlee: Currently, all HTML from Wikipedia articles are cleaned before returned, and I figured that maybe we should do the same for rendered markup which also uses Wikipedia API.

alvaromorales: Ah, you're absolutely right. I confused `clean_value` with `clean_markup`, and was wondering why we had to strip wikimarkup from an HTML value. Nvm!

alvaromorales:
lgtm :rocket:

PR #203, opened 2016-04-01
Add :file option to IMAGE-LIVE

### LINKED ISSUE #191: Add :file option to IMAGE-LIVE
The WikipediaBase 1.2.0 `IMAGE-LIVE` implementation is incompatible with START.  START will not use it, and instead will continue to use the old Omnibase implementation until it is fixed.

START wants to use `IMAGE-LIVE` to convert a specific image file _name_ to a complete image _URL_.  The current implementation of `IMAGE-LIVE` returns a complete image URL for an image file name that _it selects_ for a symbol (`IMAGE-LIVE` selects the first image returned by `IMAGE-DATA`).

`IMAGE-LIVE` should instead accept a `:file` option with the image file name:

```
(get "wikibase-term" "Tiger" (:code "IMAGE-LIVE") (:file "Tigress_at_Jim_Corbett_National_Park.jpg" :size 400))
```

START will always supply an image file name, but in the case where the file name was omitted, `IMAGE-LIVE` should select the first image returned by `IMAGE-DATA`.

[edited from @sfelshin's comment]

chsamlee:
Fixes #191.

External API changes:
- `:file` option in IMAGE-LIVE: when this option is enabled, IMAGE-LIVE returns the image with the given filename instead of the first image in article.
- if `:file` option is enabled and no file with given name is found, then a `resource-not-found` error is returned
- if the article has no images, IMAGE-LIVE will also return a `resource-not-found` error. (It used to be a generic `attribute-value-not-found` error.)

Random realization:
- there is no way for a resolver to return an empty list as result at this moment (e.g. IMAGE-DATA for an article with no images); empty lists are treated as "no result" and we continue to look for answer from other resolvers, eventually ending up with an useless `attribute-value-not-found` error.


alvaromorales:
lgtm! :rocket:

alvaromorales:
actually, could you please document `image_live`? This should be good in a comment

> :file option in IMAGE-LIVE: when this option is enabled, IMAGE-LIVE returns the image with the given filename instead of the first image in article.
if :file option is enabled and no file with given name is found, then a resource-not-found error is returned
if the article has no images, IMAGE-LIVE will also return a resource-not-found error. (It used to be a generic attribute-value-not-found error.)

Issue #87, opened 2015-10-26
Return :yyyymmdd and :html for wikibase-person


alvaromorales:
[Originally raised by @sfelshin over email]

wikipedia-person BIRTH-DATE should be less quick to return :yyyymmdd.

Note that both classes return the same result:

$ telnet nauru 8023
(get "wikipedia-person" "Sarah Bernhardt" (:code "BIRTH-DATE"))
((:yyyymmdd 18441022))

$ telnet nauru 8023
(get "wikibase-person" "Sarah Bernhardt" (:code "BIRTH-DATE"))
((:yyyymmdd 18441022))

But if we look at the Wikipedia article https://en.wikipedia.org/wiki/Sarah_Bernhardt we find

> Sarah Bernhardt (French: [sa.ʁa bɛʁ.nɑʁt];[2] c. 22/23 October 1844 – 26 March 1923) was a French stage and early film actress.

and
  
> Born	Rosine Bernardt
> c. 22 October 1844[1]
> Paris, France

It seems to me that this would be a good opportunity to return something along the lines of

$ telnet nauru 8023
(get "wikipedia-person" "Sarah Bernhardt" (:code "BIRTH-DATE"))
```
((:yyyymmdd 18441022 :html "<tr>
<th scope="row">Born</th>
<td><span class="nickname">Rosine Bernardt</span><br />
c. <span style="display:none">(<span class="bday">1844-10-22</span>)</span>22 October 1844<sup id="cite_ref-1" class="reference"><a href="#cite_note-1"><span>[</span>1<span>]</span></a></sup><br />
<span class="birthplace">Paris, France</span></td>
</tr>"))
```

$ telnet nauru 8023
(get "wikibase-person" "Sarah Bernhardt" (:code "BIRTH-DATE"))
((:yyyymmdd 18441022 :html "c. 22/23 October 1844"))

sfelshin:
See also https://github.com/infolab-csail/WikipediaBase/issues/94. In the case of Beethoven, we may question whether we should return a `:yyyymmdd` date at all, given that there is only a baptismal date and no actual birth date. If someone asks START, "Who was born earlier, Beethoven or Mozart?", it's a bit of a shame if we can't do a mathematical calculation using yyyymmdd values, but we could still fall back on giving the user each composer's non-yyyymmdd answer and letting them figure it out for themself. On the other hand, if someone asks "list composers born between 1750 and 1800", it's quite embarrassing if we leave Beethoven out.

[...]

PR #207, opened 2016-04-05
Extracting birth/death dates from paragraph (1/2)

chsamlee:
First part of the date extraction update. The code works for majority (more than 90%) of the cases, but when it doesn't work weird results can be returned. For this reason, PersonResolver's birth-date and death-date queries will continue to use information from infobox until the second part is also finished.

On the side: this PR also includes a patch to remove hidden spans from article.

Major date patterns to be implemented in the next PR:
- people born with a different name
- inexact dates
  - words before date, e.g. baptized 17 December 1770 [Beethoven]
  - phrases that indicate a period of time, e.g. early 17th century [Münejjim Bashi] (Andrew the Apostle is another example)

Update 4/21: Date queries always return `:html`. `:yyyymmdd` is returned iff the date can be represented in the format. Partially addresses #87, but the problem won't be completely fixed until PR 2.


tests/test_date_parser.py
```diff
@@ -0,0 +1,197 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
+test_date_parser
```
alvaromorales: Let's rename this file to `test_dates.py`, and move date-related tests in `test_resolvers.py` into here. I did a similar thing for image-related functionality.

Handling dates is so specific (and complex) that this deserves its own module.

tests/test_date_parser.py
```diff
@@ -0,0 +1,197 @@
...
+        self.assertEquals(d, (None, False))
+        (b, d) = get_birth_death_dates("Suzanne Scheuer")
+        self.assertEquals(b, ("February 11, 1898", True))
+        self.assertEquals(d, ("December 20, 1984", True))
+        (b, d) = get_birth_death_dates("Sergei Aleksandrovich Bershadski")
+        self.assertEquals(b, ("March 30, 1850", True))
+        self.assertEquals(d, ("1896", True))
+
+    def test_parse_symbols_interference(self):
+        # ensure that certain symbols don't interfere with date parsing
```
alvaromorales: Explain which kinds of symbols could cause interference (I don't know what you mean by "symbol")

wikipediabase/article.py
```diff
@@ -23,6 +23,12 @@ def url(self):
     def _soup(self):
         if not hasattr(self, '__soup'):
             self.__soup = fromstring(self.html_source(), title=self.title)
+            # drop hidden spans (with CSS display:none)
+            # reason for dropping: affect date parsing negatively
+            regexpNS = 'http://exslt.org/regular-expressions'
+            xpath = './/span[re:match(@style, "display:\s*none")]'
+            for e in self.__soup.xpath(xpath, namespaces={'re': regexpNS}):
+                e.drop_tree()
```
alvaromorales: Now that we merged #209, should we remove or modify these changes?

wikipediabase/date_parser.py
```diff
@@ -0,0 +1,99 @@
...
+
+# May needs no abbreviation
+MONTH_REGEX = ('((January|February|March|April|May|June|'
+               'July|August|September|October|November|December)|'
+               '(Jan|Feb|Mar|Apr|Jun|Jul|Aug|Sep|Sept|Oct|Nov|Dec)\.?)')
+DATE_REGEX = (('((\d{1,2}(/\d{1,2})* %s|%s \d{1,2}(/\d{1,2})*|%s)(,)? )?'
+               '(?P<year>\d{1,4}(\s*(/|or)\s*\d{1,4})*)'
+               '(?P<bc>( BC)?)(?P<uncertain>(( )?(\?|\(\?\)))?)')
+              % (MONTH_REGEX, MONTH_REGEX, MONTH_REGEX))
+# LOC_REGEX = '([\(\)/,\-]?( )?[^\W\d]+)+'
```
alvaromorales: This regex is commented out?

wikipediabase/date_parser.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def parse_date(raw):
```
alvaromorales: Please add a docstring for this method. Is `raw` a fragment of the first paragraph or the whole paragraph? What does the boolean in the return value mean? Also consider explaining the scoring method, and the idea of certain and uncertain dates.

It looks like you handle several date formats in one method, like:

```python
# variations on XXXX - XXXX
# (XXXX-); currently alive
# adding BC to some badly-written date range
# born XXXX / died XXXX
 # very basic tests for DoB outside parentheses
```

Have you considered refactoring each of these formats into its own method? This would make unit testing easier and more specific.

wikipediabase/date_parser.py
```diff
@@ -0,0 +1,99 @@
...
+    def clean_input(s):
```
alvaromorales: What's the goal of this method?

It shouldn't be nested inside `parse_date`, so we can add unit tests for it.

wikipediabase/date_parser.py
```diff
@@ -0,0 +1,99 @@
...
+        val = re.sub('\[(\d+|O\.S\.[^\]]+)\]', '', s)
+        val = re.sub(u'[\xa0|\s]+', ' ', val)
+        return val
+    # get phrases to run regex on
```
alvaromorales: Add a comment explaining how most first paragraphs begin with a definitional sentence like "John _is a_ scientist..."

wikipediabase/date_parser.py
```diff
@@ -0,0 +1,99 @@
...
+    best_result = ((None, False), (None, False))
+    for phrase in phrases:
+        # variations on XXXX - XXXX
+        for i in [i for i, ch in enumerate(phrase)
+                  if ch == u'\u2013' or ch == '-']:
+            birth_date, death_date = None, None
+            birth_certain, death_certain = False, False
+            birth, death = phrase[:i], phrase[i + 1:]
+            mb = re.search(DATE_REGEX, birth, re.UNICODE)
+            md = re.search(DATE_REGEX, death, re.UNICODE)
```
alvaromorales: Please use named keyword arguments (`flags=re.UNICODE`) instead of `re.UNICODE`

wikipediabase/date_parser.py
```diff
@@ -0,0 +1,99 @@
...
+                death_certain = False
+                md = True  # small hack to make condition checking easier
+            if mb and md:
+                # adding BC to some badly-written date range
+                # (e.g. 384-322 BC for Aristotle needs to be fixed)
+                if death_in_bc and not birth_in_bc:
+                    insert_idx = mb.start('bc')
+                    birth_date = (birth_date[:insert_idx] + ' BC ' +
+                                  birth_date[insert_idx:])
+                    birth_date = re.sub('  +', ' ', birth_date).strip()
```
alvaromorales: Other whitespace characters like &nbsp; values are sometimes inserted, so you should use `re.sub('\s+', ' ', flags=re.UNICODE)`

wikipediabase/resolvers/person.py
```diff
@@ -7,6 +7,7 @@
 
 from wikipediabase.article import get_article
 from wikipediabase.classifiers import InfoboxClassifier
+from wikipediabase.date_parser import parse_date
```
alvaromorales: Is `parse_date` used? If not, let's remove the import.

wikipediabase/util.py
```diff
@@ -473,3 +473,44 @@ def to_class(template):
 
     elif "taxobox" in template.lower():
         return "wikipedia-taxobox"
+
+
+def extract_parentheses_content(s, open_symbol='(', close_symbol=')', init_count=0):
```
alvaromorales: You should add unit tests in `tests/test_util.py` for this method

wikipediabase/date_parser.py
chsamlee: It is commented out because it is not used in PR 1 anymore. I am not sure if I need it in PR 2 so I'm keeping it around.

wikipediabase/date_parser.py
chsamlee: This case is not possible since we called `clean_input` in the beginning to remove other types of whitespace characters.

tests/test_frontend.py
```diff
@@ -38,8 +40,11 @@ def test_simple(self):
 
     def test_unicode(self):
         date = self.fe.eval(
-            '(get "wikipedia-military-conflict" "World War I" (:code "DATE"))')
-        self.assertEqual(date, '((:yyyymmdd 19140728))')
+            '(get "wikipedia-country" "Hong Kong" (:rendered "DEMONYM"))')
+        self.assertEqual(date,
+                         (u'((:html "<ul><li>Hongkonger</li>'
+                          u' <li>Hong Kongese</li>'
+                          u' <li>\u9999\u6e2f\u4eba</li> </ul>"))'))
```
alvaromorales: Please rename the `date` variable to something that makes more sense, e.g., `demonym`.

alvaromorales:
Only one inline comment. After that, LGTM 🚢 !

PR #209, opened 2016-04-06
Fix bugs in 'get' method and ensure clean values are returned

### LINKED ISSUE #205: 'BadRequestFormat' object has no attribute 'keys' error with invalid :markup option value
When an invalid value is passed as a `:markup` option (e.g. `foo`), the following error is raised:

```python
In [1]: from wikipediabase import wikipediabase

In [2]: wikipediabase('(get "wikipedia-artist" "Marcel Duchamp" (:rendered "Born") (:markup "foo"))')

<ipython-input-5-e263f91f514d> in <module>()
----> 1 wikipediabase('(get "wikipedia-artist" "Marcel Duchamp" (:rendered "Born") (:markup "foo"))')

code/pywikipediabase/wikipediabase/__init__.py in wikipediabase(cmd, **kw)
     16         return None
     17
---> 18     return Frontend(**kw).eval(cmd)

code/pywikipediabase/wikipediabase/frontend.py in eval(self, string)
     61         ls = loads(string)
     62         if len(ls) > 0:
---> 63             ans = self._eval(ls)
     64         else:
     65             # empty query '()'

code/pywikipediabase/wikipediabase/frontend.py in _eval(self, ls)
     44             fn = self.resources()[self._eval(ls[0])._name]
     45             args = [self._eval(a) for a in ls[1:]]
---> 46             ret = fn(*args)
     47         elif isinstance(ls[0], Keyword):
     48             # argument is a plist; convert into python dict

code/pywikipediabase/wikipediabase/knowledgebase.py in get(self, cls, symbol, attrs, options)
     28         :returns: the attribute's value(s) or an error, as a list of dict
     29         """
---> 30         return resolve(cls, symbol, attrs, **options)
     31
     32     @provide(name="get-attributes")

code/pywikipediabase/wikipediabase/resolvers/__init__.py in resolve(cls, symbol, attrs, **options)
     30     # resolvers must return a list of dictionaries
     31     for ar in get_resolvers():
---> 32         res = ar.resolve(cls, symbol, attrs, **options)
     33         if res:
     34             break

code/pywikipediabase/wikipediabase/resolvers/base.py in decorator(self, cls, *args, **kwargs)
      7     def decorator(self, cls, *args, **kwargs):
      8         if self._should_resolve(cls):
----> 9             return fn(self, cls, *args, **kwargs)
     10         else:
     11             return []

code/pywikipediabase/wikipediabase/resolvers/base.py in resolve(self, cls, symbol, attrs, **options)
     53         custom_resolve = self._custom_resolve_fn()
     54         if custom_resolve:
---> 55             return custom_resolve(cls, symbol, attrs, **options)
     56
     57         for attr in attrs.values():

code/pywikipediabase/wikipediabase/resolvers/infobox.py in resolve_infobox(self, cls, symbol, attrs, **options)
     40         answers = concat(*[ibox.get(attr, attr_type, markup=markup_option)
     41                            for ibox in infoboxes])
---> 42         answers = [self._parse_infobox_output(ans, attr) for ans in answers]
     43         # if markup is allowed: prioritize HTML answers first
     44         if markup_option == "allow":

code/pywikipediabase/wikipediabase/resolvers/infobox.py in _parse_infobox_output(obj, attr)
     51     def _parse_infobox_output(obj, attr):
     52         # don't parse markup
---> 53         if "html" not in obj.keys():
     54             return obj
     55         # HTML: parse date

AttributeError: 'BadRequestFormat' object has no attribute 'keys'
```

alvaromorales:
I looked at the output of `get` for 500+ attributes, and noticed a whole slew of bugs (proving, once again, that _Everything in Wikipedia is Hard_).

I've written detailed commit messages for each change, but here's a summary:

* Fixed bug where getting coordinates from an infobox resulted in lat/long values repeated four (4!) times
* Remove spans hidden via CSS styles because they trip up html parsing
* Only look at markup attributes with `has_value=True` in `get`. Attributes with '' or None as values caused lxml to raise an exception when attempting to parse HTML.
* Fixed bug where we choked on values with literal "<" characters
* Remove references like `[note 1]` and `[citation needed]`.
* Include all values when the markup option is set to 'allow'. Previously, we filtered answers with a 'markup' typecode in favor of answers with the 'html' typecode. The 'allow' option effectively resulted in the same behaviour as the 'skip' option, which was confusing.
* Removed the 'skip' option. Because wikimarkup syntax is pretty complex, it's actually quite hard to know whether a string contains markup or not. Our previous method only checked for templates `{{..}}` and wikilinks `[[..]]`. But even things like apostrophes (') could be wikimarkup code for boldface. In summary, it's just too hard for us to write a `is_clean_markup` function (we need a wikimarkup parser).

Fixes #205


wikipediabase/resolvers/infobox.py
```diff
@@ -40,17 +41,12 @@ def resolve_infobox(self, cls, symbol, attrs, **options):
         answers = concat(*[ibox.get(attr, attr_type, markup=markup_option)
                            for ibox in infoboxes])
         answers = [self._parse_infobox_output(ans, attr) for ans in answers]
```
chsamlee: It seems that `answers` can contain multiple errors or mix valid responses with an error. We should make sure that only the error is returned if the answers contain an error.

wikipediabase/util.py
```diff
@@ -308,13 +319,85 @@ def escape_tags(val):
...
+        (<span class="bday">1887-07-28</span>)
+      </span>28 July 1887
+    </p>
+
+    If lxml would just strip HTML tags, the value would become
+    "(1887-07-28)28 July 1887". Note the missing space between the parens -- an
+    ugly HTML value to show to the user.
+    """
+    if val.cssselect('span[style]'):
+        # unfortunately, lxml can't correctly parse nested <span> elements like
```
chsamlee: It is actually [doable](https://github.com/infolab-csail/WikipediaBase/pull/207/files#diff-6385b3d47314aae23a7d050bdc8f1c96R28) in lxml.

chsamlee:
Since we're not using `is_clean_markup` or 'skip' option anymore, we should also consider the removal of `clean_markup`. Originally, `clean_markup` was designed to be a mini wikimarkup -> HTML renderer that only handles the most common cases, while `is_clean_markup` checks if the result is human-readable. There is no point of cleaning if we're going to return possibly ugly markup or render everything. In fact, the remaining two calls of `clean_markup` are only for [removing HTML comment](https://github.com/infolab-csail/WikipediaBase/blob/markup-opt-fix/wikipediabase/infobox.py#L61) and [nothing](https://github.com/infolab-csail/WikipediaBase/blob/markup-opt-fix/wikipediabase/infobox.py#L72).

wikipediabase/util.py
chsamlee: Updated the script to fix the space issue you brought up today.
```python
regexpNS = 'http://exslt.org/regular-expressions'
xpath = './/span[re:match(@style, "display:\s*none")]'
for e in val.xpath(xpath, namespaces={'re': regexpNS}):
    e.drop_tree()
```

On when to remove hidden spans: I just remembered that both bs4 and lxml alter the HTML tree when removing elements. This feature makes calling `remove_hidden_elements` in infobox a potentially bad idea, since a client that calls Article._soup() (or any method that uses it) may get different results at different times.

alvaromorales: That's really cool, I didn't know you could write xpath expressions like that.

I tried your implementation for removing hidden `<span>`s with lxml. I also moved the call to `remove_hidden_elements` to `Article` instead of `Infobox`. However, I'm still getting a badly-cleaned answer:

```python
>>> from wikipediabase import wikipediabase
>>> wikipediabase('(get "wikipedia-artist" "Marcel Duchamp" (:rendered "Born"))')
(... (:html "(1887-07-28)28 July 1887") ...)
```

If the hidden span were properly removed, the answer would be "28 July 1887". That's why I resorted to BeautifulSoup, per my comment:

```python
    """
    WikipediaBase attempts to clean HTML before returning values from infoboxes.
    Hidden elements can cause us to produce badly-cleaned HTML. For example,
    "{{birth date|df=yes|1887|7|28}}" renders to:

    <p>
      <span style="display:none">
        (<span class="bday">1887-07-28</span>)
      </span>28 July 1887
    </p>

    If lxml would just strip HTML tags, the value would become
    "(1887-07-28)28 July 1887". Note the missing space between the parens -- an
    ugly HTML value to show to the user.

    [...]

    unfortunately, lxml can't correctly parse nested <span> elements 
    like the one shown in the docstring example
    We use BeautifulSoup instead. Note that in general we always try to
    use lxml instead of BeautifulSoup because it's significantly faster
    """
```

@chsamlee, is this just an issue with my environment? Can you reproduce?

Here's the full diff for reference:

```diff
diff --git a/wikipediabase/article.py b/wikipediabase/article.py
index 1de9e70..761531f 100644
--- a/wikipediabase/article.py
+++ b/wikipediabase/article.py
@@ -1,6 +1,7 @@
 from wikipediabase.fetcher import get_fetcher
 from wikipediabase.log import Logging
-from wikipediabase.util import Expiry, LRUCache, fromstring, tostring
+from wikipediabase.util import Expiry, LRUCache, fromstring, tostring, \
+    remove_hidden_elements
 import re


@@ -22,7 +23,9 @@ class Article(Logging):

     def _soup(self):
         if not hasattr(self, '__soup'):
-            self.__soup = fromstring(self.html_source(), title=self.title)
+            soup = fromstring(self.html_source(), title=self.title)
+            soup = remove_hidden_elements(soup)
+            self.__soup = soup

         return self.__soup

diff --git a/wikipediabase/util.py b/wikipediabase/util.py
index 91735ac..0cc110d 100644
--- a/wikipediabase/util.py
+++ b/wikipediabase/util.py
@@ -293,7 +293,6 @@ def clean_key(e_key):
 def clean_value(e_val):
     val = fromstring(unicode(e_val))
     val = handle_special_values(val)
-    val = remove_hidden_elements(val)

     val = escape_tags(tostring(val))
     val = totext(fromstring(val))  # returns the text content with no markup
@@ -381,18 +380,11 @@ def remove_hidden_elements(val):
     "(1887-07-28)28 July 1887". Note the missing space between the parens -- an
     ugly HTML value to show to the user.
     """
-    if val.cssselect('span[style]'):
-        # unfortunately, lxml can't correctly parse nested <span> elements like
-        # the one shown in the docstring example
-        # We use BeautifulSoup instead. Note that in general we always try to
-        # use lxml instead of BeautifulSoup because it's significantly faster
-        val = BeautifulSoup(tostring(val), 'lxml')
-        for elt in val.select('span[style]'):
-            for attr, value in elt.attrs.items():
-                if attr == 'style' and 'display:none' in re.sub(r'\s+', '', value).lower():
-                    # handles variations like "display:none" "display: none;"
-                    elt.extract()
-        val = fromstring(unicode(val))  # convert it back to lxml
+
+    regexp_ns = 'http://exslt.org/regular-expressions'
+    xpath = './/span[re:match(@style, "display:\s*none")]'
+    for e in val.xpath(xpath, namespaces={'re': regexp_ns}):
+        e.drop_tree()

     # remove hidden <li> elements with CSS class "mw-empty-li"
     # see https://en.wikipedia.org/wiki/MediaWiki:Common.css
```

alvaromorales:
Good point -- I removed `clean_markup`.

Less code, less bugs :bug:

wikipediabase/resolvers/infobox.py
alvaromorales: I agree. Will open an issue for improving our error handling.

wikipediabase/util.py
chsamlee: This problem can be reproduced on samoa. The problem comes from removing the line
```val = remove_hidden_elements(val)```
in util.py. Here is what happens when the query is called:

1. RenderedInfobox sees that `attr_type = 'rendered'` and tries to get the attribute value from HTML.
2. There are 3 markup attributes that map to 'born' but only 1 HTML item that has the attribute name, so HTML lookup fails.
3. We resort to rendering HTML from markup. When we send `{{birth date|df=yes|1887|7|28}}` to the Wikipedia renderer, we get HTML with hidden span.
4. `clean_value` no longer calls `remove_hidden_elements`, so we keep the hidden span in final result.

This behavior most likely originated from #198. Maybe we should reexamine the approach taken in that PR: instead of doing a markup lookup, figure out why 'assumed office' isn't in `RenderedInfobox._html_items()` (but the phrase doesn't even appear in the article!).

alvaromorales: @chsamlee, I pushed a new commit where we call the new implementation of `remove_hidden_elements` from util.py. On second thought, it makes sense to me to localize `remove_hidden_elements` only to the `get` method, instead of any other method that calls `_soup()`.

Do you think this PR is now ready to go?

> This behavior most likely originated from #198. Maybe we should reexamine the approach taken in that PR

Because of our MetaInfobox model, I actually think that doing a markup lookup is the right move, and will get us more values. ("Assumed office" renders into "In office" depending on the context).

chsamlee:
LGTM.

PR #210, opened 2016-04-09
Fix bugs and improve the request format for IMAGE-LIVE

alvaromorales:
Changed the request format for IMAGE-LIVE to `(get "wikibase-term" symbol (:code "IMAGE-LIVE") (:file "filename.jpg" :max-width 100)`.

Renamed `:size` to `:max-width`. Wikipedia's image resizing server won't resize an image if its actual size is smaller than the desired width. `:max-width` is a more descriptive name than `:size`.

The return format is a list of plists with keys `url` and `caption` (this last one is optional, and only gets included if a caption is found). Renamed `:file` to `:url` for clarity; IMAGE-DATA returns filenames but IMAGE-LIVE returns image URLs.

Example:

```
(get "wikibase-term" "Avena" (:code "IMAGE-LIVE") (:max-width 100))
((:caption "common wild oat (<i>Avena fatua</i>)" :url "https://upload.wikimedia.org/wikipedia/commons/thumb/9/9e/Avena_fatua1.jpg/100px-Avena_fatua1.jpg"))
```

Standardized the way we extract images through a util function. Improved error handling. Moved image tests to its own module.


chsamlee:
Missing tests for `clean_image_filename` and `is_valid_image_caption` is dbutil.py. Good to go otherwise.

Is there some guideline on when to use reply (vs. message) in errors?

sfelshin:
> Is there some guideline on when to use reply (vs. message) in errors?

START has a facility for "recording replies" that are tagged with "reply qualities".  After various components have had the opportunity to record replies (which usually means, after annotations have been matched against the knowledge base), START discards every reply whose quality index falls below some threshold and displays the remaining replies to the user.  I'm not sure when WikipediaBase is using "reply" vs. "message", but it would make sense if it uses "reply" for strings intended to be passed through to the user and "message" for error messages for inter-system communication.

tests/test_dbutil.py
```diff
@@ -171,3 +177,19 @@ def test_get_first_sentence_and_bolded_names(self):
...
+    def test_clean_image_filename(self):
+        fnam = '44 Bill Clinton 3x4.jpg{{!}}border '
+        clean_fnam = dbutil.clean_image_filename(fnam)
+        self.assertEqual(clean_fnam, '44_Bill_Clinton_3x4.jpg')
+
+    def is_valid_image_caption(self):
+        self.assertFalse(dbutil.is_valid_image_caption(' '))
+        self.assertFalse(dbutil.is_valid_image_caption('thumb'))
+        self.assertFalse(dbutil.is_valid_image_caption('left'))
+        self.assertTrue("Clinton's boyhood home in [[Hope, Arkansas]]]")
```
chsamlee: Should there be a call to `dbutil.is_valid_image_caption` here?

alvaromorales: Indeed! I fixed it now.

chsamlee:
LGTM!

PR #211, opened 2016-04-16
Correctly remove efn references

chsamlee:
Wikipediabase fails to remove `{{efn|...}}` references correctly because of nested curly braces. For example, running the code snippet below
```python
ibx = wikipediabase.infobox.get_infoboxes('Jesus', cls='wikipedia-person')[0]
markup = ibx.get_from_markup('birth-date')
print wikipediabase.util.remove_references_from_markup(markup)
```
yields
```
7–2 BC<!--
 -->.{{sfn|Rahner|2004|p=732}} <!--
       -->[[E. P. Sanders|Sanders]] also favors {{nowrap|c. 4 BC}} and refers to the general consensus.{{sfn|Sanders|1993|pp=10–11}} <!--
       -->[[Jack Finegan|Finegan]] uses the study of early Christian traditions to support c.&nbsp;3 or 2&nbsp;BC. <!--
 -->}}
```
which is incorrect. The correct result should be '7–2 BC' followed by empty HTML comments.


tests/test_util.py
```diff
@@ -222,6 +242,55 @@ def test_to_template(self):
                          "Template:Taxobox")
 
 
+class TestExtractParenIndices(unittest.TestCase):
```
alvaromorales: nice set of tests!

wikipediabase/util.py
```diff
@@ -445,8 +445,15 @@ def remove_references_from_markup(markup):
     markup = re.sub("<ref[^>]*>[^>]*</ref>", "", markup)  # <ref></ref>
     markup = re.sub("<ref[^>]*/>", "", markup)  # <ref/>
     markup = re.sub("\{\{reflist\}\}", "", markup)  # {{reflist}}
-    markup = re.sub("\{\{(efn|efn-ua|efn-ua|efn-lr|efn-ur|efn-lg)\|[^\}]+\}\}",
-                    "", markup)  # {{efn|...}}
+    # {{efn|...}} and its variants are special: {{...}} can nest
+    for efn in ["efn", "efn-ua", "efn-lr", "efn-ur", "efn-lg"]:
+        startIdx = markup.find("{{%s|" % efn)
```
alvaromorales: use `start_idx` instead of camelCase (`startIdx`)

wikipediabase/util.py
```diff
@@ -445,8 +445,15 @@ def remove_references_from_markup(markup):
...
+        while startIdx > -1:
+            s, e = extract_paren_indices(markup[startIdx:], "{{", "}}")[0][0]
```
alvaromorales: What if `extract_paren_indices` returns `([], 0)`? Can an `IndexError` exception be raised?

It looks like you do a `str.find` before entering the method, but I wanted to double check.

wikipediabase/util.py
```diff
@@ -537,3 +544,47 @@ def to_class(template):
...
+
+def extract_paren_indices(string, open_sym='(', close_sym=')', init_count=0):
+    """
+    Extract strings inside parentheses. Content in nested parentheses is not
+    extracted and can be found in the group of the outmost pair.
+    Parameters
+        string       : string to extract from
+        [open_sym]   : string for open parenthesis [default: '(']
+        [close_sym]  : string for closed parenthesis [default: ')']
+        [init_count] : number of extra open parentheses before s [default: 0]
```
alvaromorales: > number of extra open parentheses before s

what is `s`?

wikipediabase/util.py
```diff
@@ -537,3 +544,47 @@ def to_class(template):
...
+    Return
+        (indices, count) where
+        indices = list of (startIdx, endIdx) tuples of outermost parentheses
```
alvaromorales: change to `(start_idx, end_idx)`

wikipediabase/util.py
```diff
@@ -537,3 +544,47 @@ def to_class(template):
...
+        count = number of extra open parentheses
+    Note
+        Incomplete parentheses are included: for example, '(a)(b' yields 2
```
alvaromorales: Change "Incomplete" --> "Unbalanced" for clarity

wikipediabase/util.py
chsamlee: No exception will be raised assuming that the markup is well-formed. However, an `IndexError` exception would be raised if we call the function with `'}}{{efn'` as input.

alvaromorales: It's safer to split this up:

```python
indices, count = extract_paren_indices(markup[startIdx:], "{{", "}}")
if len(indices) > 0:
    # well-formed input, proceed as normal
else:
    # malformed input. Return `markup` as-is? Throw ValueError?
```

alvaromorales:
lgtm 🚀

PR #217, opened 2016-05-06
Fix PersonClassifier false negatives due to space

### LINKED ISSUE #216: Noam Chomsky isn't a wikibase-person
```
$ telnet kiribati 8023
(get-classes "Noam Chomsky")
("wikibase-term" "wikibase-sections" "wikipedia-philosopher")
```
This means START can't answer "How old is Noam Chomsky"—it can't match START's native annotation for "how old is _any-wikibase-person_".

chsamlee:
Fixes #216.


wikipediabase/classifiers.py
```diff
@@ -90,13 +90,13 @@ class PersonClassifier(BaseClassifier):
         ["(?P<category>.*) stubs", "\g<category>s"],
     ]
 
-    REJECT_PHRASES = [
-        "lists of ",
+    REJECT_REGEXES = [re.compile("(\A| )%s(\Z| )" % p) for p in [
```
alvaromorales: Nice catch!

With the `(\A| )` group, it looks like you're wanting to match a word boundary. Why don't you use `\b`, the special character for word boundaries? See https://docs.python.org/2/howto/regex.html#more-metacharacters

```python
 REJECT_REGEXES = [re.compile(r"\b%s\b" % p) for p in [
...
```

alvaromorales:
lgtm! 🚀

PR #221, opened 2016-05-26
Improve performance of the backend generation script and implement get-synonyms

alvaromorales:
Switched the backend generation script to `threading` because `multiprocessing` uses up too much memory. With smart caching, the script runs much faster.

Refactored various scripts used to build the WikipediaBase backend. Simplified the process and added documentation. Updated the README.

Added a `get-synonyms` method to the WikipediaBase KB.

Changed how types are loaded into the backend. We no longer lemmatize or lowercase types from whoami. We originally did so thinking that it would help with retrieval, but found that many lemmatized types were noisy. We should expect the callers (e.g., START), not WikipediaBase, to perform post-processing of types.


michaelsilver:
LGTM! 🎉

PR #223, opened 2016-05-26
Fixed print issue in lispify caused by string formatting

### LINKED ISSUE #222: Lispify cannot print
`print lispify([('How to solve inflated', [('inflaton', 'inflated'), ('too', 'to')]), ('How to solve inflation', [('inflaton', 'inflation'), ('too', 'to')])])` throws `TypeError: not all arguments converted during string formatting`

chsamlee:
Fixes #222


alvaromorales:
good catch! lgtm 🚀

PR #225, opened 2016-05-28
Fix SectionResolver behavior when markup differs from HTML

### LINKED ISSUE #220: wikibase-sections off-by-one error
`wikibase-sections` sometimes returns a following section. Example:
```
(get "wikibase-sections" "Boston" (:CODE "Government"))
((:html "<h2><span class=\"mw-headline\" id=\"Parks_and_recreation\">Parks and recreation</span></h2><div class=\"thumb tright\"><div class=\"thumbinner\" style=\"width:222px;\">...<div class=\"thumbcaption\">...
Boston Common seen from the ..."))
```
Note that the attributes are missing a section from the table of contents.  The TOC contains a "Neighborhoods" section after "Cityscapes", but this is missing from the attributes:
```
(get-attributes "wikibase-sections" "Boston")
((:code "History") (:code "Geography") (:code "Cityscapes") (:code "Climate") (:code "Demographics") ...)
```

chsamlee:
Fixes #220.


tests/test_resolvers.py
```diff
@@ -265,6 +267,60 @@ def test_sections_unknown(self):
         self.assertTrue(isinstance(err, Exception))
 
 
+class TestSectionResolversStaleMarkup(TestResolvers):
```
alvaromorales: nice testing class!

wikipediabase/resolvers/section.py
```diff
@@ -16,33 +17,106 @@ class SectionResolver(BaseResolver):
...
         markup = get_article(symbol).markup_source()
         headings = mwparserfromhell.parse(markup).filter_headings()
         ret = [h.strip("=").strip() for h in headings]
         # remove certain unwanted sections
-        unwanted = ["See also", "Related pages", "Notes and references",
-                    "Notes", "References", "Further reading", "External links"]
-        unwantedIdx = map(ret.index, filter(ret.__contains__, unwanted))
-        ret = [re.sub("\{\{anchor\|.*?\}\}", "", s).strip()
-               for s in ret[:min(unwantedIdx)]]
+        unwantedIdx = map(ret.index, filter(ret.__contains__, self.UNWANTED))
```
alvaromorales: Switch to `unwanted_idx` (same for all `camelCase` variables)

wikipediabase/resolvers/section.py
```diff
@@ -16,33 +17,106 @@ class SectionResolver(BaseResolver):
...
+                re.findall(r"[\w']+", h_heading, flags=re.UNICODE)))
+            intersection_size = len(m_words & h_words)
+            perfect_match_size = min(len(m_words), len(h_words))
+            ratio = float(intersection_size) / perfect_match_size
+            return (ratio > MIN_COMMON_RATIO)
+
+        # Treat the problem as a LCS (longest common subsequence) problem:
+        # two headings are the "same" (in the LCS sense) if the common word
+        # ratio is above a certain threshold. See the Wikipedia page on LCS
+        # for more details of the algorithm.
```
alvaromorales: This is a good and efficient solution to the problem. FYI, we're also importing the [fuzzywuzzy](https://github.com/seatgeek/fuzzywuzzy) string-matching library, that has a few ways to calculate Levenshtein Distance (we use it for fuzzy matching in infobox.py).

You may or may not find this library useful here.

alvaromorales:
Have you done random testing with enough articles to be convinced that this works for all articles, even the ones with up-to-date markup?

wikipediabase/resolvers/section.py
chsamlee: I will probably use fuzzywuzzy if the performance is acceptable. The ability to deal with typos seems to be useful when the inputs (headings) are typed by humans.

chsamlee:
I did some random testing, but not enough that I am super convinced.

chsamlee:
Done some random testing and seems to work for all articles. There are a few things that may be unacceptable though:
1. Because some sections are fetched from HTML and some are from markup, the information may be inconsistent.
2. HTML-only sections are inaccessible. For example, Boston's "cityscapes" section is split into "cityscapes" and "neighborhoods" sections. The paragraphs in "neighborhoods" cannot be accessed because `(get "wikibase-sections" "Boston" (:code "cityscapes"))` only returns the corresponding HTML section.

sfelshin:
If there is a fairly small proportion of inaccessible sections, that's not particularly troublesome as far as I'm concerned.  If there are sections for which you ask for one section but receive another, that _is_ a problem.

chsamlee:
For the most part the section received should be correct. I can think of one bad case: if the updated article removes a section called "American government" and replaces it with "Armenian government", then Wikipediabase may return the wrong section because the headings are too similar that the system thinks they talk about the same thing. But this situation is probably much rarer than unintentional typos.

sfelshin:
Okay by me, then.

alvaromorales:
lgtm 🚀

PR #226, opened 2016-05-31
Fix links with leading colons and clean up backend script

alvaromorales:
Fixed a few bugs while running the script to generate the backend. Wikimarkup ignores leading colons from links. Updated `clean_wikilink` to do so as well.

Cleaned up logging in the backend generation script. Changed the way we update counts in the backend to use less memory.


scripts/build_backend.py
```diff
@@ -131,8 +134,8 @@ def process_page_recur(page_id, title, markup, trail, is_prev_disambig,
                     article = dbutil.get_article_from_db(util.encode(l))
                     process_page_recur(article.id, util.decode(article.title),
                                        article.markup, copy.copy(trail), True, False)
-                except LookupError as e:
-                    logging.exception(e)
+                except LookupError:
+                    logging.exception("Error following link in disambig page")
```
michaelsilver: If this is logged, I don't think the person reading the log will have an easy way of knowing what disambiguation page link is being referred to. Above where you have the `Error following redirect link`, that logging statement is preceded by  `logging.debug("  .. is redirect, extracted: %s", redirect)` so the person will know which redirect, but in the `elif disambig` block, no info about the disambiguation link is ever printed. Add it either to this logging statement, or a previous one in the `elif disambig` block.

scripts/build_backend.py
```diff
@@ -403,13 +417,20 @@ def main():
...
     global types_cache, classes_cache
 
-    with futures.ThreadPoolExecutor(max_workers=min(8, args.threads)) as executor:
-        executor.map(update_counts, [(dbutil.WikiClass, value['id'], value['count'])
-                                     for value in classes_cache.values()])
-        executor.map(update_counts, [(dbutil.WikiType, value['id'], value['count'])
-                                     for value in types_cache.values()])
-        with io.open(args.output_file, mode='r', encoding='utf-8') as synonyms_file:
-            executor.map(load_synonym, synonyms_file)
+    max_workers = min(8, args.threads)
```
michaelsilver: So there will never be more than 8? Why is `args.threads` 12 by default then? A comment in the code might help.

scripts/build_backend.py
alvaromorales: Good point, but actually `logging.exception` prints a traceback that contains information about the link. Here's a snippet from the log file (the traceback shows which disambig link causes a problem):

```
Thread-3 - INFO - === evaluating: McParland
Thread-3 - DEBUG -   .. added McParland to trail
Thread-3 - DEBUG -   .. is disambig
Thread-3 - DEBUG -   .. added Anthony McParland to trail
Thread-3 - DEBUG -   .. found an article, will write
Thread-3 - DEBUG -   .. added Davie McParland to trail
Thread-3 - DEBUG -   .. found an article, will write
Thread-3 - ERROR - Error following link in disambig page
Traceback (most recent call last):
  File "./build_backend.py", line 134, in process_page_recur
    article = dbutil.get_article_from_db(util.encode(l))
  File "wikipediabase/dbutil.py", line 249, in get_article_from_db
    (title))
LookupError: Error fetching: Frank McParland. Title not found in the db
```

You can also see which disambig links get added to the trail.

I was reading about [Python logging best practices](https://www.simonmweber.com/2014/11/24/python-logging-traps.html), and learned that `logging.exception` implicitly prints out the traceback anyway—it's redundant to do `logging.exception(e)`.

michaelsilver:
LGTM

PR #227, opened 2016-05-31
Add PageRank script

alvaromorales:
Cleaned up Abdi's PageRank script. Consolidated multiple modules into a single script. Fixed style for pep8.

(cc @cheahuychou, this will be relevant when you start working on the PageRank project)


scripts/calculate_pagerank.py
```diff
@@ -0,0 +1,357 @@
...
+import io
+import multiprocessing as mp
+import re
+import time
+
+from pyspark import SparkContext
+
+from wikipediabase import dbutil
+
+# TODO: add documentation
```
chsamlee: Is this intended to be a separate issue?

scripts/calculate_pagerank.py
```diff
@@ -0,0 +1,357 @@
...
+
+    def __init__(self, in_queue, out_queue, total_rank):
+        self.in_queue = in_queue
+        self.out_queue = out_queue
+        self.total_rank = total_rank
+
+    def work(self):
+        while True:
+            key, rank = self.in_queue.get()
+            new_rank = rank  # / self.total_rank
```
chsamlee: Clarify or remove the comment. It is confusing: it seems that we should divide given that the class name has "normalize", but that isn't the case.

alvaromorales:
Thanks, I added your comment as a `TODO` code comment. I'll create an issue for documentation and the unclear commented out code. For now, I'll go ahead and merge this, but the script still needs cleaning up.

PR #233, opened 2016-06-03
Improve error handling

### LINKED ISSUE #82: Improve error handling
WikipediaBase should handle errors better.

Some suggested error types:

* **bad-request-format**: Bad request, can’t parse lisp	
* **function-not-found**: KB method doesn’t exist
* **symbol-not-found**: Symbol doesn’t exist
* **attribute-not-found**: Attribute for (symbol, class) doesn’t exist
* **class-not-found**: Class for symbol doesn’t exist
* **attribute-value-not-found**: Value doesn't exist, e.g. "currently alive"

EDIT: fixed cut&paste error

### LINKED ISSUE #204: No ability to return empty lists
In #203, @chsamlee brought up an interesting limitation (feature?) of the WikipediaBase API (slightly edited):

> There is no way for a resolver to return an **empty list** at this moment (e.g. `IMAGE-DATA` won't return an empty list for an article with no images). Empty lists are treated as "no result". We continue to look for an answer from other resolvers, eventually ending up with a useless `attribute-value-not-found` error.

@sfelshin, is this the expected behavior? Is there a situation where we would prefer an empty list over an error?

chsamlee:
Changes:
- ability to return empty list
- exceptions within Wikipediabase are now raised instead of returned as objects
- all exceptions are properly lispified
- exceptions are raised when:
  - Already present in v1.3.0:
    - query's parentheses are not balanced (`bad-request-format`)
    - query is empty (`illegal-request`, used to be `bad-request-format`)
    - image resource cannot be found (`resource-not-found`)
    - attribute cannot be found (`attribute-not-found`)
    - cannot determine value of attribute (`attribute-value-not-found`) (e.g. asking about `death-place` of a person who is still alive)
  - New in this PR:
    - query calls a nonexistent function (`function-not-found`)
    - query calls function with incorrect parameters (wrong type or wrong number of args) (`illegal-request`)
    - query asks about a symbol that doesn't exist (`symbol-not-found`)
    - the symbol specified by query does not belong to the given class (`class-not-found`)
    - for `get`, the query does not include `:code` or `:rendered` attribute (`illegal-request`)
    - query includes an invalid option (`illegal-request`, used to be `bad-request-format` for some resolvers) (e.g. `(:markup "foo")`, `(:width 0)`)
- if Wikipediabase encounters an unhandled exception, a `wikipediabase-internal-error` object is returned

Resolves #82 and fixes #204.


sfelshin:
>   - query is empty (`illegal-request`, already implemented in v1.3.0)
>   - query calls function with incorrect parameters (`illegal-request`)

Wouldn't it make sense to distinguish the error type for "query calls function with incorrect parameters"?  Maybe `invalid-arguments` or something like that?

chsamlee:
Clarification: because we have a `function-not-found` exception, it doesn't make sense to have `illegal-argument` - we already know that something is wrong with the argument.

tests/test_images.py
```diff
@@ -249,8 +250,8 @@ def test_specify_file_option(self):
                          "Map of the Warring States. Qin is shown in pink")
 
     def test_specify_nonexistent_file(self):
-        imgs = self.image_live("Boris Katz", file="random.png")
-        self.assertTrue(isinstance(imgs[0], Exception))
+        with self.assertRaises(ResourceNotFound) as e:
```
michaelsilver: Why store the exception in `e` when you are not using it?

tests/test_images.py
```diff
@@ -269,8 +270,7 @@ def test_image_live_request_format(self):
         self.assertEqual(response, expected)
 
     def test_invalid_max_width(self):
-        imgs = self.image_live('Bill Clinton', max_width='foo')
-        self.assertTrue(isinstance(imgs[0], Exception))
-
-        imgs = self.image_live('Bill Clinton', max_width=0)
-        self.assertTrue(isinstance(imgs[0], Exception))
+        with self.assertRaises(IllegalRequest) as e:
```
michaelsilver: Why store the exception in `e` when you are not using it?

tests/test_images.py
```diff
@@ -269,8 +270,7 @@ def test_image_live_request_format(self):
...
+            imgs = self.image_live('Bill Clinton', max_width='foo')
+        with self.assertRaises(IllegalRequest) as e:
```
michaelsilver: Why store the exception in `e` when you are not using it?

tests/test_resolvers.py
```diff
@@ -217,63 +222,8 @@ def test_sections_tocless(self):
...
-        err = resolve('wikibase-sections', 'Hoshizora no Memoria', {'code': 'Controversy'})[0]
-        self.assertTrue(isinstance(err, Exception))
-
-    def test_coordinates_not_found(self):
-        err = resolve('wikibase-term', 'Bill Clinton', {'code': 'coordinates'})
-        self.assertTrue(isinstance(err, list))
-        self.assertTrue(isinstance(err[0], Exception))
-
-
-class TestSectionResolvers(TestResolvers):
```
michaelsilver: Hmm... I see you've replaced some of these tests, but what about the others -- why did you delete them? (e.g. `test_sections_attributes_simple()`, `test_sections_attributes_nested()`, ...)

wikipediabase/article.py
```diff
@@ -38,21 +40,35 @@ def markup_source(self, force_live=False, expiry=Expiry.DEFAULT):
...
+                                                   force_live=force_live,
+                                                   expiry=expiry)
+            except LookupError:
+                self.exists = False
+            else:
+                self._markup_source = page
+        if self.exists:
+            return self._markup_source
+        else:
+            raise SymbolNotFound(message='Symbol not found: "%s"' % self.title)
```
michaelsilver: It's good coding practice to have one return statement not inside of an if-clause, to be sure the function returns (or raises) something no matter what. I'd suggest restructuring like this:
```python
if not self.exists:
    raise SymbolNotFound(message='Symbol not found: "%s"' % self.title)
return self._markup_source

File: wikipediabase/article.py [hunk 6]
```diff
@@ -38,21 +40,35 @@ def markup_source(self, force_live=False, expiry=Expiry.DEFAULT):
...
+            try:
+                page = get_fetcher().html_source(self.title, expiry=expiry)
+            except LookupError:
+                self.exists = False
+            else:
+                self._html_source = page
+        if self.exists:
+            return self._html_source
+        else:
+            raise SymbolNotFound(message='Symbol not found: "%s"' % self.title)
```
michaelsilver: It's good coding practice to have one return statement not inside of an if-clause, to be sure the function returns (or raises) something no matter what. I'd suggest restructuring like this:
```python
if not self.exists:
    raise SymbolNotFound(message='Symbol not found: "%s"' % self.title)
return self._html_source
```

wikipediabase/exceptions.py
```diff
@@ -0,0 +1,75 @@
...
+class IllegalRequest(WikipediabaseException):
+    """
+    The request calls functions with wrong argument(s).
+    """
+    pass
+
+
+class FunctionNotFound(WikipediabaseException):
+    """
+    The request calls a function that doesn't exist in knowledge base.
```
michaelsilver: Grammar: either say `in knowledgebase.py` if you are referring to the file or `in the knowledge base` if in the abstract. I'd prefer the latter in the case code gets refactored in the future and `knowledgebase.py` ceases to exist (not very likely, but still).

wikipediabase/exceptions.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    The value for the requested attribute does not exist. For example, asking
+    for the death date of a person who is still alive raises this exception.
+    """
+    pass
+
+
+class ResourceNotFound(WikipediabaseException):
+    """
+    The request asks for a nonexistent resource.
```
michaelsilver: Add another sentence with an example of when this is raised (when an image is not found?).

wikipediabase/frontend.py
```diff
@@ -42,7 +43,12 @@ def _eval(self, ls):
 
         if isinstance(ls[0], Symbol):
             # get the callable function
-            fn = self.resources()[self._eval(ls[0])._name]
+            name = self._eval(ls[0])._name
+            try:
+                fn = self.resources()[name]
+            except KeyError:
+                message = "Function \"%s\" not found" % name
```
michaelsilver: We strive to be Python 3 compatible with all new code we write. The `%` formatting operator is deprecated in Python 3, so use `.format()` instead. See https://pyformat.info/#simple

wikipediabase/util.py
```diff
@@ -564,3 +559,67 @@ def _normalize(x):
...
         indices.append((start, len(string)))
     return (indices, count)
+
+
+def static_checker(**type_req):
+    """
+    A decorator that checks if the right number of parameters is passed, as
+    well as if the parameters have the correct types. The decorated function
+    cannot take arbitrary keyword arguments (**kwargs). If the class of a
+    variable is not given, no type check will be performed on that variable.
```
michaelsilver: It's implicit here that one is supposed to pass the class of each variable as an argument to `static_checker()` but you should say that explicitly.

wikipediabase/util.py
```diff
@@ -564,3 +559,67 @@ def _normalize(x):
...
+
+    Raises IllegalRequest if the checker fails.
+    """
+    def decorator(fn):
```
michaelsilver: Seeing a function within a function within a function makes many python hackers' souls hurt, but I see this is sort of a special situation since you are using meta-decorators. Just add a comment with a link to http://pythoncentral.io/validate-python-function-parameters-and-return-types-with-decorators/ (which is what I presume this implementation is based on), just to excuse this seemingly odd code. It also serves as a good reference to understand this code.

wikipediabase/frontend.py
```diff
@@ -67,15 +73,19 @@ def eval(self, string):
...
+                raise exceptions.IllegalRequest(message="Empty query")
             else:
                 ans = self._eval(ls)
                 return unicode(lispify(ans))
         except Exception as e:
+            if not isinstance(e, exceptions.WikipediabaseException):
+                # bug within wikipediabase
+                cls = type(e).__name__
+                msg = cls + ': ' + e.message
+                e = exceptions.WikipediabaseInternalError(message=msg)
```
michaelsilver: Are you sure `WikipediabaseInternalError` is getting raised here (if that is your intent)? Does a unit test cover this situation?

wikipediabase/resolvers/__init__.py
```diff
@@ -28,16 +30,18 @@ def get_resolvers():
...
     # resolvers must return a list of dictionaries
+    if cls not in classify(symbol):
+        raise ClassNotFound('Cannot find class: "%s"' % cls)
     for ar in get_resolvers():
         res = ar.resolve(cls, symbol, attrs, **options)
-        if res:
-            break
-    return res
+        if res is not None:
+            return res
```
michaelsilver: Hmmm... I'm a little puzzled why you changed this
```diff
-        if res:
-            break
-    return res
+        if res is not None:
+            return res
```
What about the case when `res` is `None`? If there is a special reason you don't want anything returned, please add a comment. Otherwise, see my previous in-line comments about always having functions return something (or raise an error).

wikipediabase/resolvers/__init__.py
```diff
@@ -28,16 +30,18 @@ def get_resolvers():
...
 
 
 def resolver_attributes(cls, symbol):
+    if cls not in classify(symbol):
+        raise ClassNotFound('Cannot find class: "%s"' % cls)
     for r in get_resolvers():
         attributes = r.attributes(cls, symbol)
-        if attributes:
+        if attributes is not None:
             return attributes
```
michaelsilver: Same as above: looks like you specifically don't want the function to return anything at all in the case that `attributes` is  `None`. If this is the expected behavior and there is a specific reason for it (e.g. down the line, another function relies on it), please add a comment. Otherwise, perhaps restructure the logic so that an error gets raised (like `WikipediabaseInternalError`).

wikipediabase/resolvers/infobox.py
```diff
@@ -32,21 +33,27 @@ def resolve_infobox(self, cls, symbol, attrs, **options):
         elif "rendered" in attrs:
             attr, attr_type = attrs["rendered"], "rendered"
         else:
-            raise TypeError(("Could not resolve infobox \"%s\": must include "
-                             "code or rendered as an attribute" % (symbol)))
+            raise IllegalRequest(message='Could not resolve infobox "%s": '
+                                 'must include code or rendered as an'
+                                 'attribute' % symbol)
```
michaelsilver: We strive to be Python 3 compatible with all new code we write. The `%` formatting operator is deprecated in Python 3, so use `.format()` instead. See https://pyformat.info/#simple

wikipediabase/resolvers/infobox.py
```diff
@@ -32,21 +33,27 @@ def resolve_infobox(self, cls, symbol, attrs, **options):
...
+
         answers = concat(*[ibox.get(attr, attr_type, markup=markup_option)
                            for ibox in infoboxes])
         answers = [self._parse_infobox_output(ans, attr) for ans in answers]
-        return answers
+
+        if len(answers) > 0:
+            return answers
+        else:
+            raise AttributeValueNotFound(message="Unknown attribute value")
```
michaelsilver: It's good coding practice to have one return statement not inside of an if-clause, to be sure the function returns (or raises) something no matter what. I'd suggest restructuring like this:
```python
if len(answers) == 0:
    raise AttributeValueNotFound(message="Unknown attribute value")
return answers
```

michaelsilver:
Overall, really really nice code! Great unit tests and code organization. Making the error-handling nice and uniform like this is *huge* 🚀 . I especially like what you did with `wikipediabase/exceptions.py` and the `@static_checker` decorator -- both good calls in terms of code architecture!

Most of my comments are small and nitpicky; see the inline comments. Just to make a larger-scope point about Python 3 compatibility, even though some existing code is non-compatible, recently @alvaromorales has started using the [`__future__` import](http://stackoverflow.com/questions/7075082/what-is-future-in-python-used-for-and-how-when-to-use-it-and-how-it-works) and we hope to make the code Python 3 compatible in case we make the switch. One big reason would be for how unicode is handled, but again this would be big change and would not be likely happening soon. I didn't comment on it, but there are more `%` uses in `wikipediabase/resolvers/term.py` and `wikipediabase/resolvers/section.py` that you might want to consider addressing.

tests/test_resolvers.py
chsamlee: The SectionResolver tests are repeated for some reason (see lines 174-224 in the same file)

wikipediabase/frontend.py
chsamlee: It is my intent to raise `WikipediabaseInternalError`: if we catch an exception that is not a `WikipediabaseException`, then the exception must come from some bug in the system. Unfortunately, we can't write a unit test for these lines unless we purposely break something.

sfelshin:
Re Sam's line note

> It is my intent to raise `WikipediabaseInternalError`: if we catch an exception that is not a `WikipediabaseException`, then the exception must come from some bug in the system. Unfortunately, we can't write a unit test for these lines unless we purposely break something.

you could create a unit test by deliberately writing a broken test or deliberately adding a function to WikipediaBase that breaks.  I'm not sure what it would look like in python, but in guile I could use our [assert-throws](https://groups.csail.mit.edu/infolab/pod.cgi?infolab=share/lib/testing.scm#assert_throws_expected_expected_message_matches_thrower_lambda_message) function and write
```
(assert-throws #t "" (/ 1 0) "divide-by-zero error")
```
or alternatively
```
(define (function-that-will-throw-an-error-for-testing-purposes) (/ 1 0))
(assert-throws #t "" (function-that-will-throw-an-error-for-testing-purposes "divide-by-zero error"))
```

sfelshin:
(Note I updated my previous comment to fix a misplaced parenthesis.)

wikipediabase/frontend.py
michaelsilver: > It is my intent to raise WikipediabaseInternalError

I think (although I could be wrong) that `e = exceptions.WikipediabaseInternalError(message=msg)` is simply setting the variable `e` to be an object of `WikipediabaseInternalError`. To actually throw an error, you should instead 
```python
raise exceptions.WikipediabaseInternalError(message=msg)
```

> Unfortunately, we can't write a unit test for these lines unless we purposely break something.

@sfelshin I think @chsamlee means that this error is thrown when something unknown breaks in WikipediaBase (i.e. as far as we know, this error should never get thrown, and if it does, it's a bug in WikipediaBase), so we can't write a unit test.

chsamlee: `e = exceptions.WikipediabaseInternalError(message=msg)` replaces the unhandled error with a `WikipediabaseInternalError` object that contains information about the original. We then `lispify` the error and return the result to client. If we raise the `WikipediabaseInternalError` error instead, the client receives no response. I worked out a small script that demonstrates this:

```python
from wikipediabase.frontend import TelnetFrontend

fe = TelnetFrontend()
fe.knowledgebase.provide('bad-function', lambda: 1/0)

# should give you (:error wikipediabase-internal-error :message "XXX")
fe.eval('(bad-function)')

# serve telnet requests; run 'telnet localhost 8023' and then '(bad-function)' to see what happens
fe.run()
```

I will modify the script and put it in a unit test. The main advantage of wrapping `e` is that Wikipediabase exposes a clearly-defined error interface - the python internals won't be leaked to clients, who would see `(:error type-error)` or `(:error some-random-error-from-third-party-package)`.

wikipediabase/knowledgebase.py
```diff
@@ -8,6 +8,15 @@
 from wikipediabase.util import static_checker
 
 
+def is_list_of_dict(obj):
+    """
+    Checks whether an object is a list of dictionaries.
+    """
+    if not isinstance(obj, list):
+        return False
+    return all(map(lambda d: isinstance(d, dict), obj))
```
michaelsilver: This is a great useful function, but `util.py` seems like a better place to put it. And when you do so, please add a unit test for it :)

michaelsilver:
Once you address my one new comment, LGTM 🎉

PR #240, opened 2016-06-10
Cleanup utility functions and old files

### LINKED ISSUE #234: Clean up util functions
All helper methods in `wikipediabase/dbutil.py` should be moved to `wikipediabase/util.py` (`Markup utils` and `Image utils`). It's inconsistent to have util functions split across two different modules. 

The functions in `util.py` should be organized in a clearer way, with related functions grouped together. See `dbutil.py` and its section names, e.g., 

```python
# -------
# Section name
# -------
```

 `dbutil.py` should be renamed to `backend.py` to make it clear that the purpose of the module is to hold database modules, insert helpers, and query functions.

### LINKED ISSUE #235: Clean up old files
`CONTRIBUTING.md` and `parsetab.py` are old and outdated files that are not in use. We should delete them from the repository.

chsamlee:
Changes:
- removed CONTRIBUTING.md and parsetab.py
- removed unused functions in util.py
  - `time_interval`
  - `markup_categories`
  - `expand`
  - `url_get_dict`
  - `string_reduce`
- grouped functions into util.py into categories
- renamed dbutil.py to backend.py to make clear the purpose of module
- moved helper methods in dbutil.py to util.py
  - `clean_symbol`
  - `extract_redirect`
  - `is_disambig`
  - `extract_disambig_links`
  - `clean_wikilink`
  - `remove_local_anchor`
  - `extract_wikilinks`
  - `get_plaintext_paragraphs`
  - `clean_paragraphs`
  - `get_first_sentence_and_bolded_names`
  - `has_image_extension`
  - `extract_images_with_captions`
  - `clean_image_filename`
  - `is_valid_image_caption`
- moved tests in test_dbutil.py to util.py
- make test_resolvers.py easier to understand

Resolves #234 and resolves #235.


tests/test_util.py
```diff
@@ -150,41 +170,194 @@ def test_markup_has_valid_value(self):
...
-
-        self.assertEqual(util.clean_value(template.format('display:none;')),
-                         '28 July 1887')
-
-        self.assertEqual(util.clean_value(template.format('zoom: 100%; display: none;')),
-                         '28 July 1887')
 
-        self.assertEqual(util.clean_value(template.format('display:block')),
-                         '(1887-07-28) 28 July 1887')
+class TestMarkupUtilFromDbutil(unittest.TestCase):
```
michaelsilver: Why does the name reference dbutil? As soon as you merge the pull request, dbutil will no longer exist.

michaelsilver:
Could you explain what you did in `tests/test_resolvers.py`? You got rid of what seemed like a really elegant way of testing (subclassing `unittest.TestCase` as `TestResolvers` and having all the tests subclass `TestResolvers`). I see you wrote:
> - TODO: make test_resolvers.py easier to understand

But the way it is right now, I'm more confused because `test_resolvers.py` is halfway between two styles of testing: (1) the old way which I described above and (2) a new way which explicitly tests functions using calls with python dictionaries instead of lisp forms. If you want to commit to the new way, get rid of the `TestResolvers` class -- but I kind of like the old way as it allowed @sfelshin to say "I expect WikipediaBase to behave like x when I telnet y" and us to add a unit test for that. I'm sure you have a good reason for your changes though, so I'd be interested to know.

chsamlee:
I kept `TestResolvers` so `assertNotInDict` can be used in multiple classes, but I'm fine with moving the function out of `TestResolvers` and deleting the class. Having all tests subclass `TestResolvers` and using lisp forms causes a few problems:

- awkward syntax: `assertEqual` checks that whether a query's result matches our expectations, but to actually check whether two objects are equal, we have to do `unittest.TestCase.assertEqual(self, obj1, obj2)` which is awkward and possibly confusing.
- hard to debug: when I ran `test_resolvers.py` and found failures/errors, I had to run the entire test suite to find out the problem because it can be in any module. Because `frontend.py` wraps unhandled errors, it is not possible to see what went wrong and I have to go to the interpreter and type in the query to pinpoint the location of error. Very inconvenient. To make things worse, unittest doesn't show the difference between two strings when both are really long so I had to play the spot-the-difference game many times.

You brought up a good point that the old way allows for "I expect WikipediaBase to behave like x when I telnet y". A solution is to add another test module for integrated testing.

sfelshin:
> unittest doesn't show the difference between two strings when both are really long so I had to play the spot-the-difference game many times

You used some sort of tool to do this rather than eyeballing it, right?  E.g., in emacs, `m-x compare-windows`.

If you go with

> A solution is to add another test module for integrated testing.

then please open a new issue for this as part of closing this PR.

michaelsilver:
@chsamlee Thank you for explaining, that makes sense. Could you please move `assertNotInDict` out of `TestResolvers` and delete that class as you suggested? (You should also have a unit test for such a helper function.)

Then for readability, I'd suggest re-formatting your assertions in `test_resolvers.py` like this:
```diff
diff --git a/tests/test_resolvers.py b/tests/test_resolvers.py
index 6f5c6d2..efefaf0 100644
--- a/tests/test_resolvers.py
+++ b/tests/test_resolvers.py
@@ -61,18 +61,21 @@ class TestPersonResolvers(TestResolvers):
         self.assertEqual(e.exception.message, 'Birth/death dates not found')
 
     def test_gender_from_article(self):
-        gender = resolve('wikibase-person', 'Bill Clinton',
-                         {'code': 'gender'})
-        self.assertEqual(gender, [{'calculated': ':masculine'}])
-        gender = resolve('wikibase-person', 'William Shakespeare',
-                         {'code': 'gender'})
-        self.assertEqual(gender, [{'calculated': ':masculine'}])
-        gender = resolve('wikibase-person', 'Sacagawea',
-                         {'code': 'gender'})
-        self.assertEqual(gender, [{'calculated': ':feminine'}])
-        gender = resolve('wikibase-person', 'Mary Shakespeare',
-                         {'code': 'gender'})
-        self.assertEqual(gender, [{'calculated': ':feminine'}])
+        self.assertEqual(
+            resolve('wikibase-person', 'Bill Clinton', {'code': 'gender'}),
+            [{'calculated': ':masculine'}])
+
+        self.assertEqual(
+            resolve('wikibase-person', 'William Shakespeare', {'code': 'gender'}),
+            [{'calculated': ':masculine'}])
+
+        self.assertEqual(
+            resolve('wikibase-person', 'Sacagawea', {'code': 'gender'}),
+            [{'calculated': ':feminine'}])
+
+        self.assertEqual(
+            resolve('wikibase-person', 'Mary Shakespeare', {'code': 'gender'}),
+            [{'calculated': ':feminine'}])
         # TODO : find an example of a person who uses gender-neutral pronouns
```

I like the idea of another test module for integrated testing -- you should make an issue as Sue mentioned.

(Also don't forget my in-line comment about naming.)

chsamlee:
I found this comment in `test_resolvers.py:`
"# TODO: should DEATH-PLACE be a calculated attribute of wikibase-person?"
What is the decision on this? @sfelshin

sfelshin:
> I found this comment in `test_resolvers.py`:
> "# TODO: should DEATH-PLACE be a calculated attribute of wikibase-person?"

@chsamlee, this sounds like a question you should ask in a new issue.

michaelsilver:
LGTM

Issue #238, opened 2016-06-08
Document and add ability to flush the Redis cache when restoring the backend


michaelsilver:
(I'm adding this issue in this repository because it serves as a larger point about WikipediaBase, although it references improving a script in `github:infolab-core`.)

@cheahuychou did a `wikipediabase.article.get_article().markup_source()` on futuna and got an ancient version of the article's markup, so she asked me to update the backend. But after updating the backend from an ancient version to the most recent one using `restore-wikipediabase.sh` (and even deleting the database and trying a second time), we were still getting the same old markup.

Turns out what was happening was the Redis cache was [kicking in](https://github.com/infolab-csail/WikipediaBase/blob/master/wikipediabase/fetcher.py#L108) and returning the old cached article -- even though the backend was updated. This is deeply disturbing because when we update the backend on full START stacks for testing and/or production, we expect the answers we are getting to reflect the newest backend. But unless we flush the cache, we will continue to get old answers.

To properly address this, we need to:
- [x] Add documentation warning people about the Redis cache. This should probably go in the updated instructions on the internal infolab wiki (see https://github.com/infolab-csail/WikipediaBase/issues/104) about how to backup/restore the backend.
- [x] The last step in `github:infolab-core/bash_src/restore-wikipediabase.sh` should be a y/n question asking the user whether to clear the cache, automatically clearing it via `$ redis-cli flushall`

chsamlee:
Because redis may be used by multiple applications, we should only flush keys that are set by Wikipediabase. This involves changing `fetcher.py` and `renderer.py` to add a `wikipediabase-` prefix when getting/setting keys, and find the command that only flushes keys with specific prefix.

michaelsilver:
@chsamlee I like that idea; it seems like the most robust approach. Could you look into how complicated this is, and if it's not too hard, go ahead and implement the unique prefix? Otherwise, we can leave it for a later enhancement when we do get multiple systems using Redis.

sfelshin:
> The last step in github:infolab-core/bash_src/restore-wikipediabase.sh should be a y/n question asking the user whether to clear the cache

A couple of suggestions:
   - Rather than making this the last step, ask the question up front ("after updating the backend, should I clear the cache?", only with better wording) so that the user can execute the job, walk away, and have the job complete without further interaction.
   - Add a command line argument, and if the argument is explicitly supplied, then the user won't get asked the question.

PR #248, opened 2016-06-15
Gracefully handle Redis/PostgreSQL failures

### LINKED ISSUE #187: Gracefully handle cases when Redis and PostgreSQL are unavailable
When Redis is down or unavailable, we should just bypass the cache and succeed in returning answers. We should not raise an unintelligible error.

```
$ telnet [server] 8023
(get-classes "Bill Clinton")
((:error ConnectionError :message "Error 111 connecting to localhost:6379. Connection refused."))

$ telnet [server] 8023
(get "wikibase-term" "Bill Clinton" (:code "SHORT-ARTICLE"))
((:error ConnectionError :message "Error 111 connecting to localhost:6379. Connection refused."))
```

We should log warning statements indicating trouble connecting to redis. Additionally, we should have an infrastructure script that alerts when redis-server is down (and/or restarts it).

When postgres is down, we should probably just exit the WikipediaBase server. We can just server answers from wikipedia.org, but this is unsustainable and could lead us to getting banned.

chsamlee:
Resolves #187.

Behavior after merge:
- When Redis is down, we bypass the cache and return the answer(s). A warning (see below) is issued.
- When PostgreSQL is down, we return the error
`((:error wikipediabase-internal-error :message "PostgreSQL is unavailable"))`
- A warning includes a printed message saying which part failed and how to restart it. For production host, an email will be sent when the first error occur, and will not be sent again until the system is working correctly again. The default is to send to the user running Wikipediabse, but this behavior can be overrided by setting `WIKIPEDIABASE_ADMIN_EMAIL` environment variable.
- Add `wikipediabase:` prefix to Redis keys so we don't accidentally wipe out Redis keys used by other programs (see #238). The command to remove only Wikipediabase-related keys is
`redis-cli KEYS "wikipediabase:*" | xargs -d '\n' redis-cli DEL`.


wikipediabase/util.py
```diff
@@ -71,6 +80,47 @@ def set(self, key, value):
...
+        The same result as get_on_failure()
+    """
+    content = None
+    try:
+        content = redis_server.get(dkey)
+        if content is None or force_live:
+            content = get_on_failure()
+            # use pipeline to guarantee atomicity, because we don't
+            # want a key in Redis forever if it is supposed to expire
+            pipe = redis_server.pipeline()
```
michaelsilver: Are you sure you want to make a new pipeline for every query to redis? http://redis.io/topics/pipelining seems suggests to me it's the sort of thing where you create one pipeline and use that for all your queries.

sfelshin:
@chsamlee, I think you should alter
> For production host, an email will be sent when the first error occur, and will not be sent again until the system is working correctly again.

Instead, an email should be sent when the first error occurs, and **once daily** after that until the system is working correctly again. When things get busy around the Infolab, it's easy for the subscribers to the Start-Admins list to neglect to read digests, so daily reminders are safer.

wikipediabase/backend.py
```diff
@@ -151,45 +150,40 @@ def all_pages(only_articles=True):
...
+                       (Article.is_disambig == False) &
+                       (Article.is_redirect == False))
+                .tuples()
+            )
+        else:
+            return ServerSide(
+                Article.select(Article.id, Article.title, Article.markup)
+                .where(Article.ns == 0)
+                .tuples()
+            )
```
michaelsilver: I think it's best to keep the `result =` followed by one `return result` at the end of the function, outside of the conditional. It's good coding style.

wikipediabase/backend.py
```diff
@@ -223,15 +213,13 @@ def count_pages(only_articles=True):
...
-        result = Article.select().where(Article.ns == 0).count()
-    db.close()
-    return result
+    with open_db(db):
+        if only_articles:
+            return Article.select().where((Article.ns == 0) &  # nopep8
+                                          (Article.is_disambigx == False) &
+                                          (Article.is_redirect == False)).count()
+        else:
+            return Article.select().where(Article.ns == 0).count()
```
michaelsilver: Again, I think it's best to keep the `result =` followed by one `return result` at the end of the function, outside of the conditional. It's good coding style.

Note that where you don't have a conditional, I think it's fine to just have a one-liner `return` statement in the function. In fact, it made the code a lot clearer, so thanks for doing that 👍

wikipediabase/util.py
```diff
@@ -71,6 +80,47 @@ def set(self, key, value):
             self.cache[key] = value
 
 
+def get_from_redis(redis_server, dkey, get_on_failure, expiry, force_live):
```
michaelsilver: Could you add tests for this function in such a way that it doesn't email people? Maybe abstract away the emailing part by using a wrapper function to call this function?

wikipediabase/util.py
michaelsilver: @chsamlee and I talked and the way the code is written now actually seems fine. He's confident that the Redis pipeline is supposed to be used as it's now coded.

wikipediabase/util.py
```diff
@@ -126,6 +197,44 @@ def to_class(template):
...
+    }
+    message = messages[name]
+    print message
+    # decide whether to send an email
+    curr_host = get_system_cmd_output('hostname -f').rstrip()
+    prod_host = get_system_cmd_output('START-production-host').rstrip()
+    if curr_host == prod_host and not FAILURE_EMAIL_SENT[name]:
+        FAILURE_EMAIL_SENT[name] = True
+        addr = os.environ.get('WIKIPEDIABASE_ADMIN_EMAIL',
+                              os.environ['USER'] + '@' + curr_host)
```
michaelsilver: This will not email people's CSAIL email, but instead email the machine. If you want to email the person, just do ` os.environ['USER'] + '@csail.mit.edu'`

wikipediabase/util.py
```diff
@@ -916,3 +1025,15 @@ def is_list_of_dict(obj):
     if not isinstance(obj, list):
         return False
     return all(map(lambda d: isinstance(d, dict), obj))
+
+
+def get_system_cmd_output(cmd):
```
michaelsilver: I don't think this function is needed:
```python
>>> subprocess.check_output("hostname -f", shell=True)
'nauru.csail.mit.edu\n'
```
But see my comment above: you don't want to email `user@nauru.csail.mit.edu` but rather `user@csail.mit.edu`

wikipediabase/util.py
```diff
@@ -126,6 +197,44 @@ def to_class(template):
...
+    print message
+    # decide whether to send an email
+    curr_host = get_system_cmd_output('hostname -f').rstrip()
+    prod_host = get_system_cmd_output('START-production-host').rstrip()
+    if curr_host == prod_host and not FAILURE_EMAIL_SENT[name]:
+        FAILURE_EMAIL_SENT[name] = True
+        addr = os.environ.get('WIKIPEDIABASE_ADMIN_EMAIL',
+                              os.environ['USER'] + '@' + curr_host)
+        title = 'Service needed by Wikipediabase unavailable'
+        os.system('echo "{}" | mail -s "{}" {}'.format(message, title, addr))
```
michaelsilver: I think it'd be really nice if you could make mail be a separate function (it'll be useful in the future when we adopt python across our infrastructure code).
```python
def mail(recipients, subject, message):
    os.system('echo "{}" | mail -s "{}" {}'.format(message, subject, ','.join(recipients))
```
where `recipients` is a list

wikipediabase/util.py
sfelshin: If it emails the machine, will anyone notice?  If not, it seems rather pointless.

wikipediabase/util.py
sfelshin: Presumably this would be a function somewhere in github:infolab-core/python_src/.  I note than github:infolab-core/ruby_src/rpc.rb defines a mail function...

wikipediabase/util.py
chsamlee: I got an email in my CSAIL email when I email'ed the machine.

sfelshin: Okay, that's good, then.

michaelsilver:
👍 for using context managers for database transactions and creating a separate function for calling redis! The code is a lot simpler and cleaner, which is awesome. 

I've left some in-line comments.

wikipediabase/util.py
michaelsilver: Although putting `mail` and other commonly used functions of that sort in `github:infolab-core/python_src/` is a good long term plan, right now `github:infolab-core/python_src/` contains a bunch of caching code, and I don't want Sam to have to deal with importing all that in this PR. For now, @chsamlee just add the mail function to `util.py` in WikipediaBase, and we'll leave transferring commonly used functions and integrating `github:infolab-core/python_src/` for a later time (I'll make an issue about it).

wikipediabase/util.py
michaelsilver: Yup, just worked for me as well. Never mind what I said.

alvaromorales: WikipediaBase is an open-source project that can theoretically run on any machine. Sending emails to `@csail.mit.edu` addresses and calling the `START-production-host` restricts the project to our private infrastructure. I don't think this is the right move.

These two features can be easily implemented using environment variables (e.g., `WIKIPEDIABASE_ADMIN_EMAIL` and `WIKIPEDIABASE_SEND_EMAIL_ALERTS`).

We can then set these variables in `/infolab/share/credentials/wikipediabase.env`. Here's some (likely incorrect) Bash code showing how to set the `WIKIPEDIABASE_SEND_EMAIL_ALERTS` variable:

```bash
if [[ `whoami` == 'start' and `hostname` == `START-production-host` ]] ; then
    WIKIPEDIABASE_SEND_EMAIL_ALERTS=true
else
    WIKIPEDIABASE_SEND_EMAIL_ALERTS=false
fi
```

The README should have a "Configuration" section that documents WikipediaBase's environment variables (grep for `os.getenv` to see other variables).

If we rely on the `mail` unix utility to send emails, we should list it as a dependency on the README.

wikipediabase/util.py
```diff
@@ -80,19 +81,22 @@ def set(self, key, value):
...
-        force_live    : if true, then always use get_on_failure to get value
+        redis_server    : the Redis instance to lookup value from
+        dkey            : key used for Redis lookup
+        get_on_failure  : a zero-argument function that returns the value; this
+                          function is only called when the key is not cached
+        expiry          : how long the Redis key is valid if value is computed
+                          using get_on_failure (None for permanent key)
+        force_live      : if true, then always use get_on_failure to get value
+        [enable_warning]: if true, Wikipediabase sends warnings to users when
+                          Redis is down (default: true)
```
michaelsilver: Is `enable_warning` really needed? I feel it's redundant to simply `export WIKIPEDIABASE_SEND_EMAIL_ALERTS=false`

wikipediabase/util.py
```diff
@@ -202,9 +204,6 @@ def warn_infrastructure_failure(name):
     Warn user about unavailable infrastructure. A warning message will be
     printed for each query that uses the unavailable system. For production
     host, an email will also be sent (once).
```
michaelsilver: Update this docstring to explain how the various environment variables will affect this function's behavior.

wikipediabase/util.py
```diff
@@ -224,15 +223,24 @@ def warn_infrastructure_failure(name):
...
-                              os.environ['USER'] + '@' + curr_host)
+    email_alert_on = (os.getenv('WIKIPEDIABASE_SEND_EMAIL_ALERTS') == 'true')
+    time_since_last_email = time() - FAILURE_EMAIL_LAST_SENT[name]
+    should_send_email = (time_since_last_email > 60 * 60 * 24)  # one per day
+
+    if email_alert_on and should_send_email:
+        FAILURE_EMAIL_LAST_SENT[name] = time()
+        addresses = os.getenv('WIKIPEDIABASE_ADMIN_EMAIL')
+        if addresses is not None:
+            addresses = addresses.split(',')
```
michaelsilver: Instead of demanding that no spaces be used, you should do a [`str.strip()`](https://docs.python.org/2/library/stdtypes.html#str.strip) here to guarantee no whitespace follows or precedes the email. Be sure to get rid of the wording in the README about no spaces when you do this.

wikipediabase/util.py
```diff
@@ -1027,13 +1035,11 @@ def is_list_of_dict(obj):
...
-    Parameters:
-        cmd <str|unicode>: system command to execute
-    Returns:
-        The output of the system that would be normally printed on screen
-    """
-    proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, shell=True)
-    return proc.communicate()[0]
+# ----------
+# Infolab utils
+# ----------
```
michaelsilver: I don't like this header because `mail` is a pretty general unix utility, so it's misleading. How about just removing this header and moving `mail` immediately below the `warn_infrastructure_failure` function?

README.md
```diff
@@ -77,6 +77,21 @@ $ pip install "requests[security]"
...
+```
+To read articles from the database instead of reading from the Redis cache, set
+```bash
+WIKIPEDIABASE_DISABLE_CACHE=true
+```
+To enable email alerts when the cache and the database is down, set
+```bash
+WIKIPEDIABASE_SEND_EMAIL_ALERTS=true
+```
+To specify which email(s) to send the alerts to, set `WIKIPEDIABASE_ADMIN_EMAIL`. Use comma to separate emails and do not include space. If `WIKIPEDIABASE_SEND_EMAIL_ALERTS` is set but `WIKIPEDIABASE_ADMIN_EMAIL` is not, Wikipediabase will send emails to the machine running it.
```
michaelsilver: See my comment in `warn_infrastructure_failure` re demanding no spaces be used.

michaelsilver: Also be sure to mention here that the unix `mail` utility is required for emailing to work (i.e., when `WIKIPEDIABASE_SEND_EMAIL_ALERTS=true`).

michaelsilver:
LGTM 🎉

PR #250, opened 2016-06-17
Make get-types throw error on nonexistent article

### LINKED ISSUE #243: get-types should signal error on non-existent article
(cc @michaelsilver)
```
$ telnet navassa 8023
(get-types "one can only hope this isn't an actual article name")
()
```
Compare:
```
$ telnet navassa 8023
(get-classes "one can only hope this isn't an actual article name")
((:error LookupError :message "Error fetching: One can only hope this isn't an actual article name. Title not found in the db"))
```
On the other hand, if you think the current behavior is correct, please explain in a comment and then close the issue.

chsamlee:
Fixes #243.


wikipediabase/backend.py
```diff
@@ -297,13 +297,14 @@ def mark_as_disambig(a_id):
 def classes_of_article(article, limit=sys.maxint):
     """
     Given a case-sensitive article title, return the wikipedia classes of
-    the article (as a list of strings).
+    the article (as a list of strings). Raises a LookupError if article is not
+    found in the DB.
     """
+    article_id = get_article_from_db(article).id
```
michaelsilver: I worry about the performance cost of doing this over the original pure join. Effectively, as you have it now, there are two search queries to the database, whereas before with the pure join there was only one search query. Maybe this won't cause a performance reduction, but to be safe I'd prefer if you raised a `LookupError` in `classes_of_article()` and `types_of_article()` instead of relying on `get_article_from_db(article)`. If you have some advanced knowledge of databases and know for sure that the two approaches are equivalent, let me know -- I'd be interested in learning!

chsamlee: The problem with raising `LookupError` in `classes_of_article` and `types_of_article` is that there are 2 scenarios that return an empty list in the original code: a nonexistent article, or a real article with no classes/types (mostly redirects and disambiguation pages). I will check the performances of the old vs. new approach though - I'm not too sure either.

chsamlee:
@sfelshin There are two possible implementations for this fix. For implementation 1 (get the article and then get types from article), queries take roughly the same time regardless of the article's existence. For implementation 2 (get types as before, but check whether article exists if the returned list is empty), it is 33% faster if the article is associated with types but 33%-50% slower if it isn't (e.g. article is a redirect, disambiguation page, or doesn't exist). For START, which behavior is better on average?

michaelsilver:
> implementation 1 (get the article and then get types from article), queries take roughly the same time regardless of the article's existence

To clarify, note that although the queries take roughly the same time regardless of the article's existence, they are all about twice as slow as the original behavior when WikipediaBase simply returned an empty list regardless (before this PR). This lead us to consider
> implementation 2 (get types as before, but check whether article exists if the returned list is empty)

which has the promise of being just as fast as before, but only slower when the article has no types (or classes).

We could avoid the whole issue if we raised one error saying "either the article does not exist or it doesn't have types/classes," but I'm not sure whether that is kosher for the START end of things. The difficulty comes in when the database returns no results, and we then have to disambiguate whether that means the article is not in the database, or the article has no associated types/classes. Differentiating those two situations requires a second query to the database, causing a 2x slowdown in response time.

chsamlee:
Clarification: implementation 1 is only 1.5 times as slow as the original.

sfelshin:
I created a test case for a non-existent article for thoroughness, but it should happen rarely or never in normal START operation.  START knows what articles exist because it gets them from Omnibase's wikibase-term symbols, and Omnibase gets those from the WikipediaBase backend, which gets them from Wikipedia (eliminating redirects, disambiguation pages, and other useless articles), so START should never ask about a non-existent article.  But weird things happen sometimes.  I'm happy with either implementation 2 or Michael's "avoid the whole issue" suggestion—whichever you prefer.

chsamlee:
It doesn't make sense to slow down Wikipediabase for some cases that START never encounters, so Wikipediabase will continue to return empty lists for nonexistent articles. There will be tests on START side documenting this behavior. Feel free to reopen this issue/PR if throwing an error is necessary sometime in the future, but this is very unlikely to happen.
- from the discussion this morning

PR #252, opened 2016-06-20
Fix issues with headings in section resolver

### LINKED ISSUE #239: Section resolver raises TypeError
Query: `(get "wikibase-sections" "Homer" (:code "History and the \'\'Iliad\'\'"))`
Result: `TypeError('s2 not found')` is raised

Exact cause unknown, but probably has to do with the fact that we cannot parse the corresponding title in HTML.

chsamlee:
Fixes #239. Because Iliad is italicized, BeautifulSoup failed to get the heading from HTML. I replaced that part of the function with `Article.headings()` method which is LXML-based. Also added code to strip italics/bold from markup headings so the markup-HTML validation yields better results.


michaelsilver:
LGTM!

PR #253, opened 2016-06-21
Fix infinite loop issue with get_with_redirects

chsamlee:
`util.extract_redirect('Courtly love')` returns a section of itself when it is supposed to return `None` (the #REDIRECT is inside a HTML comment), so `backend.get_with_redirects` can't stop and won't stop. Fixed the behavior for `extract_redirect` and also added in additional safety mechanism in `get_with_redirects`.

Kudos to @cheahuychou for discovering this bug. You should be able to use get_with_redirects instead of get_article_from_db without any problem after this PR is merged.

Also fixes 2 broken test cases with Bill Clinton.


wikipediabase/util.py
```diff
@@ -568,7 +568,7 @@ def clean_symbol(symbol):
 
 
 def extract_redirect(markup):
-    redirect_match = re.search(REDIRECT_PATTERN, markup)
+    redirect_match = re.match(REDIRECT_PATTERN, markup)
```
michaelsilver: Just out of curiosity, why did you decide to change `search` to `match` (which [only matches at the beginning of the string](https://docs.python.org/2/library/re.html#search-vs-match))?

michaelsilver:
Good catch and nice solution! LGTM

chsamlee:
It's nicer than throwing an `'\A'` at the beginning of `REDIRECT_PATTERN`. The real redirects' #REDIRECT are always at the beginning, so changing from `search` to `match` is the easiest solution.

PR #254, opened 2016-06-22
Fix MetaInfobox generation in backend script

chsamlee:
In the previous backend, the log file contains many errors and no meta infoboxes are added to the backend. The cause is that we did not define WIKIBASE_RENDERER in [renderer.get_renderer()](https://github.com/infolab-csail/WikipediaBase/blob/51ae306f6c80d5b498519f00365912bf2cd176c5/wikipediabase/renderer.py#L87) before logging a warning, which yielded the "'NoneType' object has no attribute 'log'" errors.

(Edit: Ignore the Travis failures caused by Bill Clinton. They're dealt with in the other PR.)


michaelsilver:
Ugh 😭  #NullHandlingErrorKillsA5DayLongScript

LGTM

PR #255, opened 2016-06-22
Fix typo in backend.count_pages

chsamlee:


michaelsilver:
The change looks good but I'm a little disturbed by the number of unit tests that are currently failing. Could you look into that before merging?

PR #256, opened 2016-06-25
Tests for backend.py module

chsamlee:
Added `test_backend.py` module. The module builds a test backend with ~10 articles and runs tests with the test backend without affecting the normal backend. The module can correctly report the 2 backend-related bugs that we recently found (missing synonyms, and no meta-infoboxes in DB).

This test module will be skipped locally by default. To enable it, set env variable `WIKIPEDIABASE_TEST_BACKEND=true`. This is done for 2 reasons:

1. It seems that Wikipediabase has 2 flavors: one with backend and one without. The `build_backend.py` script requires additional dependencies. If the ability to build backend is not included by default, then testing of backend function should not be included by default either because one of its most important function is to test the correctness of build script.
2. It takes more time to run `test_backend.py` than to run all the other tests combined.

Note: `test_backend.py` always runs on Travis.

On the side:
- `configure_backend.py` and `test_backend.py` now take a `-n` option that indicates the name of the DB in which the backend will be built. If this parameter is not present, the default option ("wikipediabase") is used.
- It is now possible to tell Wikipediabase to use a different backend by calling `backend.use_db`. The backend used at startup is the current default ("wikipediabase"). One implication is that we can switch to a backup backend if the primary fails, though we probably don't need this functionality anytime soon.

(P.S. This is PR number 2^8 and the 101st PR. Too bad PR number 2^8-1 aka the 100th PR is a typo fix patch :/)


scripts/configure_backend.py
```diff
@@ -66,19 +67,29 @@ def create_database_and_role(name='wikipediabase'):
     print "Created role"
 
 
-def create_tables():
-    db.connect()
-    create_model_tables(Model.__subclasses__(), fail_silently=True)
+def create_tables(name, drop):
```
michaelsilver: Even though you have default values in the `ArgumentParser` below, I think it'll be even cleaner/clearer code to have the default values hardcoded into the functions, i.e., `create_tables(name='wikipediabase', drop=False)`. (This is also nice to have in case another program wants to call these functions all on their own.)

scripts/configure_backend.py
```diff
@@ -48,7 +49,7 @@ def check_psql_command(cmd):
     check_output(['psql', '-U', 'postgres', '-h', 'localhost', '-c', cmd])
 
 
-def create_database_and_role(name='wikipediabase'):
+def create_database_and_role(name):
```
michaelsilver: Why get rid of the default value here? See my comment below.

michaelsilver:
Wow, this is really amazing work, @chsamlee! I love it 🎉 

The only thing I'm not a huge fan of is the 10min Travis build time. Luckily, I think we can dramatically speed this up by [caching pip install dependencies](https://docs.travis-ci.com/user/caching/#pip-cache). This was previously removed in PR https://github.com/infolab-csail/WikipediaBase/pull/76 due to concerns described in issue https://github.com/infolab-csail/WikipediaBase/issues/75. I think a good compromise is to only cache the hefty `script-requirements.txt` like this:
```diff
diff --git a/.travis.yml b/.travis.yml
index 1820f0c..4c970ee 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 sudo: false
 language: python
+cache:
+  directories:
+    - $HOME/.cache/pip
 python:
     - "2.7"
 services:
@@ -11,7 +14,7 @@ install:
     - python -m nltk.downloader punkt | head -n 1000
     - python -m nltk.downloader wordnet | head -n 1000
     - python setup.py install
-    - pip install -r scripts/script-requirements.txt
+    - pip install -r scripts/script-requirements.txt --download-cache $HOME/.cache/pip
     - python -m spacy.en.download
 before_script:
     - scripts/configure_backend.py
```

michaelsilver:
On second thought, it looks like my hacky way of caching doesn't just cache the `scripts/script-requirements.txt`, but it actually caches all the dependencies. We get a [depreciation error](https://travis-ci.org/infolab-csail/WikipediaBase/builds/140707671#L692) which is not great. The good news is that it looks like the way pip caching on travis works, it respects specific version numbers, and lets the setup.py or requirements.txt files guide what is installed (so if you delete a dependency that is needed, it won't get installed, and you'll see your code stop working). This resolves the issue I brought up in #75. 

@chsamlee can you undo the change I suggested in the `.travis.yml` and instead just add `cache: pip` right under the `language: python` line? (see https://docs.travis-ci.com/user/caching/#pip-cache). This is the standard way of pip caching on travis.

michaelsilver:
LGTM (but before you merge, just fix the one test that is failing due to a change on Wikipedia).

Looks like we're still stuck at a 10 min build mainly due to the many heavy dependencies that require `gcc` compilation and other serious build steps, so caching doesn't work for them (for example, `spacy`). We might want to return to this in the future, but for now I think we can live with a 10 min build.

scripts/configure_backend.py
sfelshin: I probably am misunderstanding due to my ignorance about Python, but isn't it better _not_ to hardwire default values?

If I was programming this in Common Lisp, I'd set up a global variable to hold the default value for the option, and initialize it to an initial, global default value.  I'd also make the relevant function accept an override value for the option.  When the code is executed, the caller could either let the initial, global default value stand, change the value of the global variable and let that default value prevail, or explicitly pass in a value to the function which would override the global variable regardless of whether the global variable was currently bound to the initial value or to some new value.  (For extra jollies, I might create a second global variable to hold the same initial, global default value, so that my program could see whether the global variable was set to the initial default or some new default.)

But maybe Python does things differently, or maybe that's what you're already doing and I don't recognize it.

michaelsilver: > When the code is executed, the caller could either let the initial, global default value stand, change the value of the global variable and let that default value prevail, or explicitly pass in a value to the function which would override the global variable regardless of whether the global variable was currently bound to the initial value or to some new value. 

This is sort of how it works in python, except the variables are local in scope.

A normal function like this:
```python
def create_tables(name, drop):
```
requires two arguments. But when you add default values (which I called "hardcoding", but that's not really a good way to describe it), all of a sudden, those arguments become optional:
```python
def create_tables(name="some default", drop=False):
```
in that case you can call the function in four different ways:
```python
create_tables() # uses the both default values for name and drop
create_tables(name="different name") # changes the name only, but keeps the value of drop (which is False)
create_tables(drop=True) # change the value of drop to True, but keeps name as "some default"
create_tables("different name", True) # or equivalently:
create_tables(name="different name", drop=True) # which redefines both the name and drop arguments
```
Ref: http://docs.python-guide.org/en/latest/writing/style/?#function-arguments

These two references might be helpful for learning more about python in general if you are interested:
  * [Python for Lisp Programmers](http://norvig.com/python-lisp.html)
  * [Python for Programmers with 3 Hours](https://stephensugden.com/crash_into_python/)

sfelshin: Thanks.  I was confused by "hardcoded", but now I get it.

PR #259, opened 2016-06-29
Improve Travis build time

chsamlee:
Create 2 jobs for each Travis build: 1 with a backend and 1 without. That way the one without the backend can give feedback in 5 minutes instead of 10.


michaelsilver:
LGTM

PR #262, opened 2016-07-12
Strip leading and trailing spaces in redirects

chsamlee:


michaelsilver:
Could you add a unit test for the article that caused the code to trip up before you added the one line to strip whitespace?

tests/test_util.py
```diff
@@ -202,6 +202,11 @@ def test_redirect_with_leading_colon(self):
         self.assertEquals('Jelgava–Liepāja Railway',
                           util.extract_redirect(markup))
 
+    def test_redirect_with_trailing_space(self):
+        markup = "#REDIRECT [[Everybody's Got to Learn Sometime ]]"
```
michaelsilver: Because "Everything in Wikipedia is hard" ™️ ©️ ®️ and we want to be robust,  why don't you add a space in front of the article, in case someone does that somewhere on Wikipedia. Or, do you think that is illegal markup, so it would never happen?

chsamlee: I think that is illegal markup, but I can't find a source stating that. But then I can't find any rule stating that leading and trailing spaces are allowed either.

michaelsilver: Why not just add a leading space, update the unit test name accordingly, and then move on comforted that you'll probably never encounter this bug again, even if a leading space might never actually happen IRL?

michaelsilver:
LGTM!

PR #264, opened 2016-07-15
Add other variables to pagerank scoring

cheahuychou:
Included other characteristics of each Wikipedia page in rank calculation such as article length and whether the page is protected, featured and good. 

Enabled rank calculation using different weights

Cleaned up and simply the original code so it runs faster and easier to use. For example, there are
default `wikifile` and `symbols_to_lengths` files, so one does not need to regenerate those files if only running PageRank

Reimplemented `create_wikifile` so that the disambiguation and namespace and invalid links are filtered out and the redirects are replaced by the actual symbols as the file is being created which should save a lot time since there is no postprocessing needed

Attempted the page views in the scoring, however there were not any APIs out there that we could use to get the average page views within certain time frame in a scalable time (e.g. https://tools.wmflabs.org/pageviews/)


scripts/calculate_pagerank.py
```diff
@@ -1,548 +1,496 @@
 #!/usr/bin/env python
 
 """
-Calculates PageRank scores for pages in Wikipedia.
+Calculates PageRank scores for pages in Wikipedia. The scores are calculated based on
+the specified weights for the scores from PageRank algorithm and other characteristics
+of the page such as its approximnate and whether it is a featured, semi-protected, and
```
michaelsilver: Typo: `approximnate`

scripts/calculate_pagerank.py
```diff
@@ -1,548 +1,496 @@
...
+good page.
 
 Depends on pySpark. To install:
 1. Download Spark from http://www.apache.org/dyn/closer.lua/spark/spark-1.6.1/spark-1.6.1.tgz
 2. To build Spark, first install Scala build tool (sbt) by following the instructions on http://www.scala-sbt.org/release/docs/Installing-sbt-on-Linux.html under "Ubuntu and other Debian-based distributions"
 3. Unzip the .tgz file, cd into the folder and run `sbt assembly` to build Spark
-4. To run pySpark, do `bin/pyspark`
+4. To confirm that the build was successfully, do `bin/pyspark` to run PySpark
```
michaelsilver: Typo: `successfully` -> `successful`

scripts/calculate_pagerank.py
```diff
@@ -1,548 +1,496 @@
...
 
 DELIMETER = '<LINKDELIM>'
-INCOMPLETED = '<INCOMPLETED>'
-global semi_protected_symbols, featured_symbols
+SEMI_PROTECTED_SYMBOLS, FEATURED_SYMBOLS, GOOD_SYMBOLS = None, None, None
+ALL_SYMBOLS, ALL_REDIRECTS, ALL_DISAMB = {}, {}, set([])
+MAX_ARTICLE_LENGTH, SYMBOLS_LENGTHS = None, {}
+MAX_PAGE_RANK = 20000  # based on the experiments
+WIKIFILE_PATH = os.getenv('WIKIFILE_PATH')
+LENGTH_PATH = os.getenv('LENGTH_PATH')
```
michaelsilver: Per the README in `/infolab/share/credentials/`
> Environment variables should be defined in *.env files. Variables should be
prefixed with the system name (e.g. OMNIBASE_) to avoid name conflicts.

Prefix these variables with `WIKIPEDIABASE_PAGERANK_`, and add them to `/infolab/share/credentials/wikipediabase.env`

scripts/calculate_pagerank.py
```diff
@@ -1,548 +1,496 @@
...
+def main(args):
+    wikifile_path = args.wikifile_path
+    processed_path = args.processed_path
+    iterations = args.iterations
+    num_pages = args.num_pages
+    pagerank_weight = args.pagerank
+    featured_weight = args.featured
+    good_weight = args.good
+    protected_weight = args.protected
+    length_weight = args.length
```
michaelsilver: Could you make it so that `main()` accepts a series of arguments: `wikifile_path, processed_path, etc.` and then move the calls to `args.` to the `if __name__ == "__main__":` part of the script? That way, `main(wikifile_path, processed_path, ...)` could be called from another script.

scripts/calculate_pagerank.py
```diff
@@ -1,548 +1,496 @@
...
+
+def pagerank(wikifile_path, processed_path, pagerank_weight, other_weights, iterations=5, num_pages=None):
     """
-    Returns the approximate word count of section, does not exclude headings
-    and references etc.
+    Runs PageRank
+
+    params:
+    - wikifile_path: path to the file that contains that contains lines of the pattern:
+                     page_title DELIMITER page_id [DELIMITER wikilink]*
```
michaelsilver: What kind of delimiter? Could you also clarify what you mean by a wikilink?

scripts/calculate_pagerank.py
```diff
@@ -1,548 +1,496 @@
...
+                     for all wikipedia symbols
+    - processed_path: path to file that will have final PageRank results
+    - iterations: number of iterations to perform, needs to be sufficiently
+                  many to make the ranks converge
+    - num_pages: number of pages to rank
```
michaelsilver: what about `other_weights`?

scripts/calculate_pagerank.py
```diff
@@ -1,548 +1,496 @@
...
+    rank = pagerank_weight * original_rank / MAX_PAGE_RANK
+    if is_featured(page):
+        rank += featured_weight
+    if is_good(page):
+        rank += good_weight
+    if is_protected(page):
+        rank += protected_weight
+    if page in SYMBOLS_LENGTHS:
+        rank += length_weight * \
+            min(1, float(SYMBOLS_LENGTHS[page]) / MAX_ARTICLE_LENGTH)
```
michaelsilver: Its interesting that you chose to use additive weights for additional features. I wonder what would happen if you multiplied by the weight? (Not needed for this PR, just food for thought.)

scripts/calculate_pagerank.py
```diff
@@ -1,548 +1,496 @@
...
+    with open_db(db):
+        iterable = ServerSide(Article.select(Article.title)
+                              .where((Article.ns == 0) &
+                                     (Article.is_disambig is False) &
+                                     (Article.is_redirect is False))
+                              .tuples())
+    global ALL_SYMBOLS
+    for elt in iterable:
+        symbol = elt[0].strip()
+        ALL_SYMBOLS[symbol] = None
```
michaelsilver: You could probably achieve this faster using a list comprehension:
```python
ALL_SYMBOLS = {elt[0].strip(): None for elt in iterable}
```

scripts/calculate_pagerank.py
```diff
@@ -1,548 +1,496 @@
...
+                    protected.add(actual_symbol.encode(
+                        "ascii", "xmlcharrefreplace"))
+            if "Category:Wikipedia_semi-protected_pages" in url:
+                next_page_elt = soup.findAll(
+                    attrs={'title': 'Category:Wikipedia semi-protected pages'})[-1]
+            elif "Category:Wikipedia_indefinitely_semi-protected_pages" in url:
+                next_page_elt = soup.findAll(
+                    attrs={'title': 'Category:Wikipedia indefinitely semi-protected pages'})[-1]
+            if next_page_elt.getText() == 'next page':
+                url = 'https://en.wikipedia.org' + next_page_elt['href']
```
michaelsilver: Looks like you are going to be making repeated requests to wikipedia -- do you have random delays between calls? This is really important so that we don't get banned from `wikipedia.org`. (See also https://github.com/infolab-csail/infolab-core/issues/44).

michaelsilver:
I'm a little bugged by the heavy use of global variables... any particular reason why you chose to go this route?

scripts/calculate_pagerank.py
cheahuychou: I have already tried that, please see previous commits. That did not work because the difference was not very noticeable and the ranking did not change that much.

scripts/calculate_pagerank.py
cheahuychou: It shouldn't be an issue there since we are only making about 20 requests.

scripts/calculate_pagerank.py
cheahuychou: How will `main` get called from another script? the name `main` isn't very useful.

scripts/calculate_pagerank.py
cheahuychou: I believe that wikilink is a term used WikipediaBase, at least in `util.py`. There is also an explanation in the spec for `create_wikifile`.  Just didn't want to repeat it, could include it up here too.

cheahuychou:
The global variables are the constants that are used by various functions in the order that makes it quite troublesome and confusing to be passed around. I have got rid of the paths and the delimiter. Unless you have any suggestions, I will keep the rest.

scripts/calculate_pagerank.py
michaelsilver: Although certainly the more requests that are made, the higher the chance of getting banned, the fact that a script makes requests in rapid succession could get us banned, even at 20 requests. All it takes is getting banned once, and potentially all of WIkipediaBase will be in trouble, which is a very serious matter. It should only slow down your script a little bit, so please take the appropriate precaution.

scripts/calculate_pagerank.py
michaelsilver: You're right; `main()` is actually a pretty useless function. Why don't you just put all of it inside of `if __name__ == "__main__":`, and then that way, `pagerank` will be the topmost function.

When you do this, don't copy over
```python
    wikifile_path = args.wikifile_path
    processed_path = args.processed_path
    iterations = args.iterations
    num_pages = args.num_pages
    pagerank_weight = args.pagerank
    featured_weight = args.featured
    good_weight = args.good
    protected_weight = args.protected
    length_weight = args.length
```
just use `args.length` directly instead of `length_weight`.

cheahuychou: So move everything in `main(args)` to `if __name__ == "__main__":` and get rid of `main`? that makes much more sense to me

scripts/calculate_pagerank.py
michaelsilver: Can you make`other_weights` explicit arguments instead of a list? Or, alternatively, remove `pagerank_weight` and have one argument `weights` that is a list of all the weights. (I think the latter is better; feels more mathematical, like having a vector of weights.)

scripts/calculate_pagerank.py
michaelsilver: Yes, exactly.

scripts/calculate_pagerank.py
michaelsilver: Hm, that's sort of surprising. I'd expect a multiplicative scalar to have much more of an effect than addition. But there are many ways to skin the cat -- as long we get good rankings at the end, it doesn't matter :)

michaelsilver:
LGTM

PR #267, opened 2016-07-18
Fix sections classifier behavior

### LINKED ISSUE #266: Is get-classes supposed to include wikibase-sections when there aren't any?
(FYI, navassa is running WikipediaBase 1.3._n_.)
```
$ telnet navassa 8023
(get-classes "Dongbaekseom")
("wikibase-term" "wikibase-sections")
$ telnet navassa 8023
(get-attributes "wikibase-sections" "Dongbaekseom")
()
```
It's okay if this is intentional behavior, but if it is, it means that _every_ article belongs to _both_ `wikibase-term` and `wikibase-sections`, and that should be documented as correct behavior.

In case you're wondering, I noticed this because I was trying to find an article that has coordinates but has no infobox and doesn't belong to `wikibase-sections`, in order to test a corner case related to coordinates. https://en.wikipedia.org/wiki/Dongbaekseom has neither infobox nor sections, so I was surprised to see that it has class `wikibase-sections`.

chsamlee:
classify() should now include wikibase-sections iff the article includes
a real section (which excludes sections like "References")

Fixes #266


wikipediabase/classifiers.py
```diff
@@ -40,8 +40,34 @@ def classify(self, symbol):
...
+                title = heading.group('title')
+                return (title not in self.UNWANTED_SECTIONS)
+        return False
+
     def classify(self, symbol):
-        return ['wikibase-sections']
+        classes = []
+        if self.has_sections(symbol):
+            classes += ['wikibase-sections']
+        return classes
```
michaelsilver: Maybe I'm missing something here, but why cant this function just be:

```python
if self.has_sections(symbol):
    return ['wikibase-sections']
return []
```

chsamlee: Good point. I copied this from PersonClassifier, but have no idea why this is done in the first place.

michaelsilver:
Is `UNWANTED_SECTIONS` the only redundancy between the classifiers and resolvers, or are there other things you copied over? If so, what are your thoughts on creating a file where such things are shared between the two? I fear the code debt we might get ourselves into for introducing such redundancies (such as maintaining two separate lists of the same thing).

wikipediabase/classifiers.py
michaelsilver: Lol, such things happen. Just be sure to update all places with such unnecessary list concatenation.

chsamlee:
`UNWANTED_SECTIONS` is the only redundancy. Compared to creating a file, importing one from the other is probably a cleaner solution.

michaelsilver:
> Compared to creating a file, importing one from the other is probably a cleaner solution.

Good point! Why don't you do that?

michaelsilver:
LGTM (but make sure travis passes first)!

PR #269, opened 2016-07-19
Fix get_with_redirects errors due to encoding

chsamlee:
Wikipediabase stores article titles using URL encoding, but get_article_from_db requires titles in HTML encoding. This causes badly-decoded redirects that eventually yields DoesNotExist errors.

Edit: apparently both unicode and percent-encoded string are acceptable for wikimarkup. I hate wikimarkup's lax standards.

P.S. This is a temporary fix. Once Allegro fixes its unicode issue, we should standardize how unicode strings are stored in Wikipediabase.


sfelshin:
Allegro _can_ handle Unicode, but since START gets Wikipedia article titles through Omnibase, [START is blocked](https://github.com/infolab-csail/START/issues/71) until [Omnibase can handle Unicode](https://github.com/infolab-csail/omnibase/issues/7), which is waiting on [switching from Guile to Chicken](https://github.com/infolab-csail/omnibase/issues/79), which is well underway but a big change.

michaelsilver:
LGTM!

> This is a temporary fix. Once Allegro fixes its unicode issue, we should standardize how unicode strings are stored in Wikipediabase.

Indeed (but see Sue's response -- it's not an issue in Alegro, it an issue with Guile). If there isn't an  issue for this, why not go ahead and create one, with a reference to this PR so we don't forget.

michaelsilver:
@sfelshin just to confirm, was there no issue back when Alvaro was working on WikipediaBase relating to the way Allegro decodes utf-8 strings? The WikipediaBase codebase is littered with comments like this one:
```python
# As of Oct '15, Allegro Lisp doesn't correctly decode a utf-8 string.
# As a temporary workaround, we encode into numeric HTML entities
# e.g. "&#160;" for Unicode non-breaking space
# TODO: remove when unicode issue in Allegro is fixed
```

sfelshin:
@michaelsilver, there was and still is an issue. My point is purely that the comment you reference above:
```
# TODO: remove when unicode issue in Allegro is fixed
```
inaccurately blames the Allegro implementation of Common Lisp and should instead place blame something like this:
```
# TODO: remove when unicode issue in START (due to Omnibase, due to Guile 1.8) is fixed
```
and likewise for similar comments.  (Can be trimmed to `TODO: remove when unicode issue in START/Omnibase is fixed` if a terser comment is preferred.)

There already is an [issue about this for WikipediaBase](https://github.com/infolab-csail/WikipediaBase/issues/86). Good idea  :sparkles: to cross-reference it and this PR (which I have just done) since the issue is rather old and neglected, and the changes to WikipediaBase for Unicode may have drifted.

PR #271, opened 2016-08-08
Fix pagerank bug and print progress

cheahuychou:
- Correct `sections` which was misspelled earlier
- Fix a unicode bug in `extract_valid_wikilinks`
- Add a timeout wrapper around `get_with_redirects` since despite many bug fixes, the function still got stuck at a few redirects which didn't turn out to be necessary 
- Add some print statements to show progress of `load_redirects` since that function takes quite a long while to finish


chsamlee:
I'm not sure if we should use timeouts to stop `get_with_redirects`. The function either keeps running after timeout, or terminates without properly closing the database connection - but both cases lead to resource leak.

I still think stopping `get_with_redirects` after a certain number of redirects is a better idea. A good number to stop is 20, which also reduces the maximum runtime of `get_with_redirects` to 1 second. Additionally, failures should be logged in Wikipediabase so we can investigate later.

cheahuychou:
@chsamlee If you could make `get_with_redirects` do that, that will be great. I can then undo that change.

chsamlee:
LGTM, but you should wait for Travis tests to pass.

PR #272, opened 2016-08-08
Limited redirects and fix tests

chsamlee:
`backend.get_with_redirects` now take an optional argument `max_redirects`. If `max_redirects` is specified, the `(max_redirects + 1)`th redirect will throw a LookupError. Infinite looping is still the default.


chsamlee:
(Ignore the Jim Guy Tucker vs. Frank D. White test case. I know the reason why it fails, but fixing it will take a while.)

wikipediabase/backend.py
```diff
@@ -279,7 +279,7 @@ def get_meta_infobox_from_db(template):
                               'the db'.format(template))
 
 
-def get_with_redirects(symbol):
+def get_with_redirects(symbol, max_redirects=-1):
```
michaelsilver: Could you please update the docstring?

sfelshin:
> (Ignore the Jim Guy Tucker vs. Frank D. White test case. I know the reason why it fails, but fixing it will take a while.)

Should you mark it as "expected to fail" and open a new issue for it (with whatever priority you deem fit), so that it neither interferes with test cases passing nor gets lost in the shuffle?

michaelsilver:
LGTM! 🎉 , but please first wait for Travis to pass.

alvaromorales:
@chsamlee, this feels like an incomplete solution to the problem of circular redirects, as the value of `max_redirects` is arbitrary. What if `get_with_redirects(symbol)` calls a helper function `get_with_redirects_recur(symbol, visited)`, where `visited` is a set of pages that have already been followed? This helper should check if `symbol` is in `visited`. If so, this means we've hit a circular redirect and we should throw an exception.

In this way, we'll stop processing as soon as we hit the first circular redirect and won't need to keep going `max_redirect` times.

PR #274, opened 2016-08-09
Use visited pages to detect circular redirects

chsamlee:
`max_redirects` is arbitrary and not user-friendly as it's hard to know what value it should be set to. We use visited pages to detect circular redirects as soon as one appears instead of waiting for `max_redirect` redirects to finish.

The type signature of `get_with_redirects` is reverted to before #272, and a `LookupError` is raised if circular redirect is detected.


wikipediabase/backend.py
```diff
@@ -279,32 +279,24 @@ def get_meta_infobox_from_db(template):
...
     Raises a LookupError if the symbol or any of its redirects are not found
-    in the DB, or if the redirect is too deep
+    in the DB, or if the redirects are circular
     """
+    visited = set([symbol])
     article = get_article_from_db(symbol)
     redirect = extract_redirect(article.markup)
-    prev_redirect = None
-    redirect_count = 0
+    visited.add(redirect)
```
michaelsilver: To avoid repeating stuff outside and inside of the while loop just for initialization purposes, I'd suggest implementing a [loop and a half](https://users.cs.duke.edu/~ola/patterns/plopd/loops.html#loop-and-a-half).

wikipediabase/backend.py
```diff
@@ -279,32 +279,23 @@ def get_meta_infobox_from_db(template):
...
-        prev_redirect = redirect
+    visited = set()
+    article, redirect = None, symbol
+    while True:
+        visited.add(redirect)
         article = get_article_from_db(redirect)
         redirect = extract_redirect(article.markup)
-        if redirect_count == max_redirects:
-            raise LookupError('Error fetching: {}. Too many redirects.'
+        if redirect is None:
```
michaelsilver: What about the `redirect != prev_redirect` logic?

michaelsilver: Oops, right, never mind -- your check for circular redirects is even stronger.

michaelsilver:
LGTM!

wikipediabase/backend.py
chsamlee: The `redirect != prev_redirect` logic is introduced in #253 as an "additional safety mechanism" to exit an infinite redirect, but it doesn't seem to work well. Whether we should return a 1-article redirect loop (which is the only case handled by the logic) is another question.

wikipediabase/backend.py
```diff
@@ -279,32 +279,23 @@ def get_meta_infobox_from_db(template):
...
+            break
+        if redirect in visited:
+            raise LookupError('Error fetching: {}. Circular redirects.'
```
michaelsilver: After some in person discussion @chsamlee agrees: since in theory this should never happen, it should be a `WikipediabaseInternalError`

wikipediabase/backend.py
michaelsilver: I think throwing an error is better since in theory there shouldn't be infinite redirects in Wikipedia, and if there is a case of infinite redirects, we should know about it.

michaelsilver:
LGTM, for real this time :)

wikipediabase/backend.py
```diff
@@ -279,32 +280,26 @@ def get_meta_infobox_from_db(template):
...
         redirect = extract_redirect(article.markup)
-        if redirect_count == max_redirects:
-            raise LookupError('Error fetching: {}. Too many redirects.'
-                              .format(symbol))
-        else:
-            redirect_count += 1
+        if redirect is None:
+            break
+        if redirect in visited:
+            # There shouldn't be any circular redirects in wikipedia as a page with
```
alvaromorales: This isn't true, circular redirects are everywhere in Wikipedia! We ran into similar infinite loop issues when we first wrote the build_backend.py script.

Redirects are created by humans, just like regular pages, so you can see why it's easy to inadvertently create a circular loop. My best guess is that the Mediawiki software has a way to check for circular redirects so that pages aren't unloadable. 

While it's possible that there might be (unrelated) bugs in util.extract_redirects, I'm fairly confident that this isn't causing circular loops.

I don't think it's accurate for the exception to be a WikipediabaseInternalError, as this is caused by bad data in Wikipedia.

PR #279, opened 2016-10-21
Ignore articles starting with "Capital of "

sfelshin:
Ignore all articles starting with "Capital of " because they cause
subsumption problems for START.  E.g., START was failing to answer "How
far is the capital of China from the capital of Sudan?" because "Capital
of China" redirects to the article for "Beijing".  (Someday we'll make
Omnibase do something better about subsumption; see
https://github.com/infolab-csail/omnibase/issues/68.)

Closes https://github.com/infolab-csail/WikipediaBase/issues/278.


michaelsilver:
@amirfarhat although the change @sfelshin made in this Pull Request is trivial and doesn't break anything, it looks like some unit tests are failing due to changes to articles on Wikipedia. 

Take a look at the [test failures](https://travis-ci.org/infolab-csail/WikipediaBase/jobs/169591581#L1084) and read about unit testing [here](http://docs.python-guide.org/en/latest/writing/tests/) (only the intro and unittest section) and [here](https://docs.python.org/2/library/unittest.html).

Let's find a time offline to meet in person to peer program this!

michaelsilver:
@amirfarhat since we didn't quite get to peer program fixing some of these unit tests, here's a write up of what I would have gone through: 

# Big picture
In general in software programming, failing unit tests generally means there's a bug in your code.

However, since WikipediaBase uses live data in many of it's tests, half of the time unit tests fail due to Wikipedia changing. 

To determine whether it's a real bug or just a unit test that needs to be updated to reflect a change in Wikipedia, you look at the documentation of the function under testing.

# One example
## Why is our test failing?
Let's now work through fixing [one failing unit test](https://travis-ci.org/infolab-csail/WikipediaBase/jobs/169591299#L1079-L1082)

We see:
```python
File "/home/travis/build/infolab-csail/WikipediaBase/tests/test_infobox.py", line 273, in test_tag_annotation_stripping
    self.assertIn("Paul Murphy", ibox.values())
```

So we look in [`WikipediaBase/tests/test_infobox.py: 273`](https://github.com/infolab-csail/WikipediaBase/blob/master/tests/test_infobox.py#L273)

Then we figure out what the heck [`get_rendered_infoboxes()`](https://github.com/infolab-csail/WikipediaBase/blob/master/tests/test_infobox.py#L25) does.

So we look in `wikipediabase.infobox` and [find this](https://github.com/infolab-csail/WikipediaBase/blob/master/wikipediabase/infobox.py#L1079-L1097).

Ah! `get_rendered_infoboxes()` returns a list of [`RenderedInfobox` objects](https://github.com/infolab-csail/WikipediaBase/blob/master/wikipediabase/infobox.py#L171).

Ok, and what's our goal again? To figure out what's going on in:
```python
def test_tag_annotation_stripping(self):
    ibox = dict(get_rendered_infoboxes("Manchester")[0]._html_items())
    self.assertEqual(ibox["GDP"], 'US$ 92.3 billion')
    self.assertIn("Paul Murphy", ibox.values())
```

Ah, right. So we figure out that [`_html_items()` returns a list of key : value pairs for the HTML attributes of that infobox](https://github.com/infolab-csail/WikipediaBase/blob/master/wikipediabase/infobox.py#L283-L287).

So we go and look at the [Manchester Wikipedia page]( https://en.wikipedia.org/wiki/Manchester) and notice that "Paul Murphy" is not in the infobox. Darn, Wikipedia changed. Now we have to find another appropriate example for the unit test.

## What are we testing?
The name of the unit test suggests we're testing `._html_values()`'s ability to remove HTML tags. But we can consult the git history on the file [by clicking "blame"](https://github.com/infolab-csail/WikipediaBase/blame/master/tests/test_infobox.py#L273) to see what commit added that unit test.
- [Commit](https://github.com/infolab-csail/WikipediaBase/commit/fcc21fe1bdca811277788dda5ca534cac6cd55eb)
- [Corresponding pull request](https://github.com/infolab-csail/WikipediaBase/pull/149) (small "#149") at top of previous page
- [Corresponding issue](https://github.com/infolab-csail/WikipediaBase/pull/105)

So looks like this change was made in the context of removing HMTL tags from the infobox values, and the particular commit that added this unit tests adds the ability to remove superscript/subscript HTML tags.

So let's look at what the [Manchester page used to look like](https://en.wikipedia.org/w/index.php?title=Manchester&oldid=737751920)

(I found that by clicking "View History" at the top of the Manchester Wiki page).

Ah and viewing source in the browser and searching for "Paul Murphey" leads us to find:
```HTML
<th scope="row">&#160;•&#160;<a href="/wiki/Lord_Mayor_of_Manchester" class="mw-redirect" title="Lord Mayor of Manchester">Lord Mayor</a></th><td>Paul Murphy<sup id="cite_ref-1" class="reference"><a href="#cite_note-1">[1]</a></sup></td>
```
Notice the `<sup>` ... `</sup>` tags. On the actual page itself that's why Paul Murphey has a small "[1]".

Ok. Alright. Now we know we're hunting for an example of a Wikipedia infobox where the value has a superscript. 

## Updating the unit test
And hey! We don't have to look far: https://en.wikipedia.org/wiki/Manchester

has "Carl Austin-Behan" as the new mayor! With a superscript!

So literally all you have to do is change the unit test:
```python
self.assertIn("Paul Murphy", ibox.values())
```
to
```python
self.assertIn("Carl Austin-Behan", ibox.values())
```

This way we are sure that the unit test still tests the same ability in the code (to remove the superscript tag in an infobox attribute).

# Actionables
- Check out this branch (`capital-of`), enter your virtualenv, and run the unit tests:
```
$ git checkout capital-of
$ workon wikibase
(wikibase)$ python setup.py test
```
- Go through each unit test similar to how I did in the example above. 
- Update the tests as needed, commit and push to the `capital-of` branch.
- Assign the PR to me to review.

alvaromorales:
LGTM

PR #287, opened 2017-05-09
Clarify README.md

korrawat:
Add an instruction for installing dependencies.


README.md
```diff
@@ -21,9 +21,10 @@ Install `pip` and `redis-server` if not installed:
 
 > Caution: Until wikipedibase is deployed on START installing it for non-testing purposes is not something one is supposed to do.
 
-Install `wikipediabase`:
+Install `wikipediabase` and dependencies:
 
     python setup.py install
+    pip install -r requirements.txt
```
michaelsilver: Swap the order of these steps. 

1) `pip install -r requirements.txt` (installs specific versions of the dependencies)

2) `python setup.py install` (installs any version of the dependencies)

If you do (2) before (1), (1) will uninstall the wrong versions installed by (2) and re-install the correct versions of the dependencies.

Instead, do (1) first to from the start get the correct versions of the dependencies, and then (2) to install WikipediaBase as a Python module.

michaelsilver:
LGTM!

Travis is going to fail because the tests are out of date. See https://github.com/infolab-csail/WikipediaBase/issues/288

Issue #288, opened 2017-05-09
Update tests to reflect new Wikipedia data


michaelsilver:
Wikipedia has changed, but we haven't updated the tests in a while. So, lots of tests are failing.

Go through the failures one by one and update the test cases with new ones based on the new data on Wikipedia. A worked example on how to do this can be found [here](https://github.com/infolab-csail/WikipediaBase/pull/279#issuecomment-257988060).

Issue #273, opened 2016-08-08
Infobox.get returns wrong answer due to ordering


chsamlee:
Query:
```python
ibx = wikipediabase.infobox.get_rendered_infoboxes('Bill Clinton')[0]
# expects "Frank D. White" as answer, but returns "Jim Guy Tucker"
print ibx.get('successor2', 'code')
```

In the current code, we assume that [the ordering of items is kept when markup is rendered into HTML](https://github.com/infolab-csail/WikipediaBase/blob/d0999b89fbada6b669884055f334d92a2b97f32f/wikipediabase/infobox.py#L242-L245), which is apparently not the case for Bill Clinton. In `ibx._markup_items()`, White comes before Tucker; but in `ibx._html_items()`, Tucker comes before White.

A simple solution is to get the markup value and render the result, instead of trying to guess which HTML item is the correct one. Not sure how much impact this has on speed, and the person implementing this change should ensure that `get` is still efficient after the change.

PR #289, opened 2017-05-16
Fix failing test cases to reflect changes in Wikipedia

korrawat:
Fixed 19 failures and 1 error in response to #288 

There are two test cases in `test_infobox.py` that I'm not sure what they are testing and thus have not fixed. 

- [`test_get_code_conflicting_rendered_name`](https://github.com/infolab-csail/WikipediaBase/blob/fixup/tests/test_infobox.py#L88) - This test is marked `expectedFailure` in #272, but now the test actually passes. I'm not sure why it was expected to fail then.
- [`test_rendered_attributes_no_footnotes`](https://github.com/infolab-csail/WikipediaBase/blob/fixup/tests/test_infobox.py#L324) - This test is shown below. Currently, `ibx.attributes['qs-a']` is just `QS`, but not `'QS \(Global version\)QS \(Asia version\)'` or `'QS \(Global version\) QS \(Asia version\)'` as before.
```python
    def test_rendered_attributes_no_footnotes(self):
        ibx = MetaInfoboxBuilder('Template:Infobox world university ranking').build()
        self.assertEqual(ibx.attributes['arwu-a'], 'ARWU')
        self.assertRegexpMatches(ibx.attributes['qs-a'],
                                 'QS \(Global version\) ?QS \(Asia version\)')
```

Fix https://github.com/infolab-csail/WikipediaBase/issues/288


chsamlee:
First test case: See #273.
Second test case: It tests that the attribute isn't rendered as "QS[34]" or something like that, so changing the regex to "QS" should be fine.

korrawat:
@chsamlee Thank you!

First test case: the test case no longer fails because the Wikipedia source code has been changed. The ordering of items in markup now matches the HTML, i.e. `successor1 = [[Jim Guy Tucker]]` now comes before `successor2 = Frank D. White` (see #273). Does that mean I can just remove `@unittest.expectedFailure`? Or should I find another Wikipedia article that demonstrates the same bug? (What does "conflicting" in `test_get_code_conflicting_rendered_name` mean?)

Second test case: Why is `'QS \(Global version\) ?QS \(Asia version\)'` changed into just `QS`? Does this result from the change in the Wikipedia article (but I see very little change in the article from that time to now) or the change in Wikipediabase implementation?

chsamlee:
First test case: You can remove the decorator since finding another article with the same bug is going to take a long time. I believe that the name "conflicting" means that multiple markup attributes correspond to the same HTML attribute: both"successor2" and "successor3" are rendered as "successor"; both "birth-date" and "birth-place" are rendered as "born".

Second test case: The change results from changes in templates. In this case, it is a change in `{{{QS_A}}}` entry in "Template:Infobox world university ranking/Asia" that causes the difference.

alvaromorales:
@korrawat your changes look good. Travis is failing because of an old spaCy version. We need to wait on https://github.com/infolab-csail/whoami/issues/75 before updating the pip dependencies in `scripts/script-requirements.txt`. Once that's fixed, this PR can be merged.

pondpremt:
The command for downloading spaCy data changes to `python -m spacy download en`. I'll fix that on the whoami repo, but I believe it has to be changed here also?

sfelshin:
> I'll merge this to unblock WikipediaBase development.

🎉 🌟 💖 ❗️

